PR #1303, opened 2021-07-29
Changes to support make validate-runenv-image targets using podman

paulpopelka:
Change the validate-runenv-image make target to also run the container image using podman.
Pick up a small krun fix to allow the java payload's runenv-image to work with podman.
Added a .check_podman_prereqs make target that verifies prerequisites needed for podman are done.

With these changes the following payloads validate-runenv-image targets work:
busybox, node, python, java

### DIFF
--- a/container-runtime/podman_config.sh
+++ b/container-runtime/podman_config.sh
@@ -15,24 +15,27 @@
 # limitations under the License.
 #
 
-
-#
 # A little script to get needed linux packages, change podman config files, and add a small
 # kontain selinux policy to allow podman to run containers using krun and km.
-#
+
+TRACE=x
+
+# exit if any command fails, don't execute commands if TRACE has a value.
+set -e ; [ "$TRACE" ] && set -x
+
+# Should be run as root.
+[ `id -u` != "0" ] && echo "Must run as root" && exit 1
 
 # podman config file locations
-CONTAINERS_CONF=/usr/share/containers/containers.conf
-HOME_CONTAINERS_CONF=~/.config/containers/containers.conf
-ETC_CONTAINERS_CONF=/etc/containers/containers.conf
 DOCKER_INIT=/usr/libexec/docker/docker-init
 KRUN_PATH=/opt/kontain/bin/krun
 KM_PATH=/opt/kontain/bin/km
 KM_SELINUX_CONTEXT="system_u:object_r:bin_t:s0"
-
-# Either do it or just print the commands we think should be run.
-#DEBUG=echo
-DEBUG=""
+CONTAINERS_CONF=/usr/share/containers/containers.conf
+ETC_CONTAINERS_CONF=/etc/containers/containers.conf
+# if running under sudo, update the invokers containers.conf, not root's
+# Note that the shell nazi's think eval to expand ~ is evil but I can't find a better way
+HOME_CONTAINERS_CONF=`eval echo ~${SUDO_USER}/.config/containers/containers.conf`
 
 linuxdist=`grep "^ID=" /etc/os-release`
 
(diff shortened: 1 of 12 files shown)


container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
```
sv641, line 1: license

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+      echo "Existing init_path:"
+      grep "^ *init_path" $HOME_CONTAINERS_CONF
+   else
+      echo "init_path = \"$DOCKER_INIT\"" is already in $HOME_CONTAINERS_CONF
+   fi
+fi
+
+
+# add krun to ~/.config/containers/containers.conf at the beginning of the [engine.runtimes] section
+echo
```
sv641, line 56: Is there a reason for a bunch of empty echos..

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+   echo "selinux-policy-devel is not installed, installing it"
+   $DEBUG sudo dnf -y install selinux-policy-devel
+else
+   echo "selinux-policy-devel is already installed"
+fi
+
+
+# Add kontain selinux policy adjustments
+POLDIR=/tmp/kontain_selinux_policy
+if ! test -d $POLDIR
```
sv641, line 89: [[ ! -d $POLDIR ]] &&  mkdir -p $POLDIR 

Is bash idiomatic.

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+then
+   mkdir $POLDIR
```
sv641, line 91: always add -m to have correct permissions.

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+allow container_t kvm_device_t:chr_file { append getattr ioctl lock open read write };
+
+# You may get errors when this is "compiled".
+# This is a known issue, see:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1861968
+EOF
+
+if ! test -e Makefile
+then
+   $DEBUG ln -s /usr/share/selinux/devel/Makefile
```
sv641, line 112: this fails when DEBUG is enabled.

#DEBUG=echo
DEBUG=""

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+CONTAINERS_CONF=/usr/share/containers/containers.conf
+HOME_CONTAINERS_CONF=~/.config/containers/containers.conf
+ETC_CONTAINERS_CONF=/etc/containers/containers.conf
+DOCKER_INIT=/usr/libexec/docker/docker-init
+KRUN_PATH=/opt/kontain/bin/krun
+KM_PATH=/opt/kontain/bin/km
+KM_SELINUX_CONTEXT="system_u:object_r:bin_t:s0"
+
+# Either do it or just print the commands we think should be run.
+#DEBUG=echo
```
sv641, line 17: rest of the code will either work or not based on what DEBUG is set to.
if DEBUG is set to echo then the code wort work. Is it to just get trace info?

sv641: Is there a reason not to use bash debug model "-x"

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+DEBUG=""
+
+# Install podman
+if ! dnf list installed | grep -q podman
+then
+   echo "podman is not installed, installing it"
+   $DEBUG sudo dnf -y install podman
+else
+   echo "podman is installed"
+fi
```
gnode1, line 27: Why not just install it? It will do nothing if already installed, and say nothing with -q `sudo dnf -q install podman`

gnode1, line 27: Second question - dnf implies RedHat system. What about Ubuntu?

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# set init_path in the [containers] section of ~/.config/containers/containers.conf
+# We don't ensure we add this in the [containers] section.
+# We hope the init_path command line exists and is in the proper section.
+echo
+if ! grep -q "^ *init_path" $HOME_CONTAINERS_CONF
+then
+   # no init_path, use the docker init
+   $DEBUG sed -e "/^# *init_path/ainit_path = \"$DOCKER_INIT\"" $HOME_CONTAINERS_CONF >/tmp/sed$$.out
+   if test $? -eq 0
```
gnode1, line 38: sed has `--in-place` option, makes this unnecessary

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+
+
+# add krun to ~/.config/containers/containers.conf at the beginning of the [engine.runtimes] section
+echo
+if ! grep -q "^ *krun =" $HOME_CONTAINERS_CONF
+then
+   $DEBUG sed -e "/^\[engine.runtimes\]/akrun = [\n   \"$KRUN_PATH\",\n]\n" $HOME_CONTAINERS_CONF >/tmp/sed$$.out
+   if test $? -eq 0
+   then
+      mv /tmp/sed$$.out $HOME_CONTAINERS_CONF
```
gnode1, line 62: again `--in-place` for sed

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+
+
+# set selinux context on /opt/kontain/bin/km
+echo
+$DEBUG chcon $KM_SELINUX_CONTEXT $KM_PATH
+
+
+# Get policy build tools.
+echo
+if ! dnf list installed | grep -q selinux-policy-devel
```
gnode1, line 78: same as above dnf comment

container-runtime/podman_config.sh
gnode1, line 89: Or better yet `mkdir -p`

container-runtime/podman_config.sh
```diff
@@ -0,0 +1,117 @@
...
+}
+
+allow container_t kvm_device_t:chr_file { append getattr ioctl lock open read write };
+
+# You may get errors when this is "compiled".
+# This is a known issue, see:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1861968
+EOF
+
+if ! test -e Makefile
```
gnode1, line 110: Can the Makefile come from anywhere else? If not `ln -sf` would just do the same in one line

make/images.mk
```diff
@@ -173,6 +173,10 @@ validate-runenv-image: ## Validate runtime image
 	${KM_DOCKER_VOLUME} \
 	${RUNENV_DEMO_IMG_TAGGED} \
 	"$${tmp_bash_array[@]}" | grep "${RUNENV_VALIDATE_EXPECTED}"
+	tmp_bash_array=${RUNENV_VALIDATE_CMD} && \
+	${PODMAN_RUN_TEST} \
+	docker-daemon:${RUNENV_DEMO_IMG_TAGGED} \
```
gnode1, line 178: I don't understand the `docker-daemon` here

paulpopelka: podman has its own local image repository that it will use by default.
We need to tell podman to get the image from the docker daemon's local image repositry.

paulpopelka:
Closing this PR since its changes have been incorporated into the PR for serge/validate-runenv-2:
PR #1322
https://github.com/kontainapp/km/pull/1322

PR #1310, opened 2021-08-05
makefiles cleanup

gnode1:
Fix makefiles so subdirs traverse works as intended. Please note - no make target should be mentioned before `include actions.mk` in a Makefile.

Cleaned up docker volumes mounted for payload builds. One of the advantages - payloads now can bu build and tested in parallel: `make -C payload -j` builds them, `make -C payloads -j runenv-images` and so on. The only little gotcha is that for validate-runenv-images need to say `make validate-runenv-image -j DOCKER_INTERACTIVE=` - parallel make disconnect tty for parallel runs.


paulpopelka:
> Please note - no make target should be mentioned before `include actions.mk` in a Makefile.

The above should be added as a comment in actions.mk.

gnode1:
> > Please note - no make target should be mentioned before `include actions.mk` in a Makefile.
> 
> The above should be added as a comment in actions.mk.

Done

payloads/python/Makefile
```diff
@@ -75,10 +75,9 @@ endef
 all:
 ifneq ($(FROMSRC),yes)
 	${DOCKER_RUN_BUILD} \
-		-v ${TOP}:${TOP}:Z \
-		-v ${KM_OPT_RT}:${KM_OPT_RT}:Z \
-		-v ${KM_OPT_BIN}:${KM_OPT_BIN}:Z \
-		-v ${KM_OPT_LIB}:${KM_OPT_LIB}:Z \
+		-v ${TOP}:${TOP}:z \
+		${KM_OPT_VOLUME} \
+		-v ${KM_OPT_LIB}:${KM_OPT_LIB}:z \
```
paulpopelka, line 80: Not needed.  KM_OPT_VOLUME already has KM_OPT_LIB in it.

gnode1: Good catch, thanks!

paulpopelka (approved):

PR #1315, opened 2021-08-10
Preserve SA_RESTORER address in snapshots

johnmuth81:
Snapshots were not preserving the SA_RESTORER flag but not the address.
This caused signal handlers created with the SA_RESTORER flag to attempt
to return to 0. We found this in the SpringBoot demo, but it effected more
than that.

The test excersizes the condition. Verified that the test fails without
the changes to KM.

### DIFF
--- a/tests/snapshot_test.c
+++ b/tests/snapshot_test.c
@@ -37,6 +37,10 @@
 
 #include "km_hcalls.h"
 
+#ifndef SA_RESTORER
+#define SA_RESTORER   0x04000000
+#endif
+
 char* cmdname = "???";
 int do_abort = 0;
 int do_stdclose = 0;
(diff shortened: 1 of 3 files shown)


tests/snapshot_test.c
```diff
@@ -95,6 +107,12 @@ void setup_process_state()
    CHECK_SYSCALL(epollfd = epoll_create(1));
    // leave a gap in FS space for recovery
    CHECK_SYSCALL(close(tmpfd));
+
+   struct sigaction act = {
+       .sa_sigaction = test_sigaction,
+       .sa_flags = SA_SIGINFO,
+   };
+   CHECK_SYSCALL(sigaction(SIGUSR1, &act, NULL));
    return;
```
paulpopelka, line 116: Could you add some code here to pull the sigaction structure back from the kernel and verify that SA_RESTORER has been set by the c library?

paulpopelka (commented):
Looks fine with one comment.

tests/snapshot_test.c
johnmuth81: done

paulpopelka (approved):

gnode1 (approved):

PR #1316, opened 2021-08-12
automate release publish

gnode1:
Build KM, KKM, and runenv, smoke test, then build and publish artifacts. The artifacts are Vagrant boxes on Vagrant cloud, AMI, and runenvs on dockerhub.

The release is triggered by pushing the tag like v*.*.*


johnmuth81 (approved):
lgtm

.github/workflows/km-ci-workflow.yaml
```diff
@@ -114,7 +115,7 @@ jobs:
       - run: make -C payloads push-testenv-image
 
       - name: Create payloads runenv and demo-runenv images
-        run: make -C payloads runenv-image demo-runenv-image
+        run: make -C payloads runenv-image
 
       - run: make -C payloads push-demo-runenv-image
```
paulpopelka, line 121: This looks weird.
We say we are building demo-runenv-image but we aren't anymore, and we are pushing the demo-runenv-image we are no longer building.

gnode1: `make runenv-image` now (for a while actually) builds both runenv and demo-runenv images. Maybe a comment would help?

.github/workflows/km-ci-workflow.yaml
```diff
@@ -340,7 +340,7 @@ jobs:
     if: (failure() && github.ref == 'refs/heads/master') ||
       contains(github.workflow, 'noisy')
     # Dependencies. (A skipped dependency is considered satisfied)
-    needs: [km-test, kkm-test, kkm-test-aws, km-test-all]
+    needs: [km-build, kkm-build, km-test, km-test-withpacker, kkm-test, kkm-test-aws, km-test-all]
```
paulpopelka, line 344: Does this job only report fails in the jobs listed here?
It doesn't not follow the dependencies between jobs to decide what to report?

Makefile
```diff
@@ -37,13 +37,14 @@ ifeq ($(shell uname), Darwin)
 MAKE := make
 endif
 
-kkm-pkg: ## Build KKM module self-extracting package.
+.kkm-pkg:
+	makeself -q kkm ${BLDTOP}/kkm.run "beta-release" ./installer/build-script.sh
+
+kkm-pkg: .kkm-pkg ## Build KKM module self-extracting package.
 	@if ! type makeself >& /dev/null ; then echo Please install \"makeself\" first; false; fi
 	@# We only needs source code, so running 'clean' first
```
paulpopelka, line 45: We only need ....
not needs.

Makefile
```diff
@@ -37,13 +37,14 @@ ifeq ($(shell uname), Darwin)
...
 	$(MAKE) MAKEFLAGS="$(MAKEFLAGS)" -C ${TOP}/kkm/kkm clean
```
paulpopelka, line 46: You've changed the order of the steps here.
Before we ran make and then makeself,
now we run makeself and then make.
Does the makeself depend on what was produced by the make invocation?

.github/workflows/km-ci-workflow.yaml
gnode1: I'm not sure about that. It says "A skipped dependency is considered satisfied". I thought this way it is more explicit

Makefile
gnode1: good catch. However make does clean, so it is not producing anything, it is removing things. Still, need to restore the order

make/actions.mk
```diff
@@ -64,11 +64,12 @@ push-testenv-image: subdirs ## Push test images to a cloud registry. IMAGE_VERSI
 pull-testenv-image: subdirs ## Pulls test images from a cloud registry. IMAGE_VERSION is mandatory. Used mainly in CI
 runenv-image: subdirs ## builds and packages image with runtime environment for each payload
 push-runenv-image: subdirs ## Push runtime images to a cloud registry. IMAGE_VERSION can be used used to modify tag name
+push-demo-runenv-image: subdirs ## push runtime demo images to cloud registry with multiple release tags. IMAGE_VERSION can be used used to modify tag name
 pull-runenv-image: subdirs ## Pulls runtime images from a cloud registry. IMAGE_VERSION an be used used to modify tag name
+pull-demo-runenv-image: subdirs ## Pulls runtime test images from a cloud registry. IMAGE_VERSION an be used used to modify tag name
```
paulpopelka, line 69: "can be used" not "an be used".

"used used" should be "used"

make/actions.mk
```diff
@@ -64,11 +64,12 @@ push-testenv-image: subdirs ## Push test images to a cloud registry. IMAGE_VERSI
 pull-testenv-image: subdirs ## Pulls test images from a cloud registry. IMAGE_VERSION is mandatory. Used mainly in CI
 runenv-image: subdirs ## builds and packages image with runtime environment for each payload
 push-runenv-image: subdirs ## Push runtime images to a cloud registry. IMAGE_VERSION can be used used to modify tag name
+push-demo-runenv-image: subdirs ## push runtime demo images to cloud registry with multiple release tags. IMAGE_VERSION can be used used to modify tag name
```
paulpopelka, line 67: "used used" should be just "used"

make/locations.mk
```diff
@@ -14,8 +14,8 @@
 # limitations under the License.
 #
 #
-# A helper include. ALlowes to include some vars in dirs which cannot include
-# actions.mk (e.g. tests which need their own compike /link flags)
+# A helper include. Allowes to include some vars in dirs which cannot include
```
paulpopelka, line 17: Allowed not Allowes?

paulpopelka (approved):

Issue #1124, opened 2021-04-13
validate Krun / KM with podman


msterin:

paulpopelka:
The work to fix this was done with the following change:
947175f Changes to support make validate-*-image targets using podman (#1322)
https://github.com/kontainapp/km/pull/1322

PR #1322, opened 2021-08-23
Changes to support make validate-*-image targets using podman

paulpopelka:
Change the validate-runenv-image make target to also run the container image using podman.
Pick up a small krun fix to allow the java payload's runenv-image to work with podman.
Added a .check_podman_prereqs make target that verifies prerequisites needed for podman done.
Added a podman_config.sh script to make podman config changes and selinux policy changes so that podman run works
with krun and km.
Added docker_config.sh scrip to make docker config changes so that docker run will use krun.
Change to fetch podman from the proper ubuntu package repository.
Tell apt-get it is ok to install gnupg2
Try apg-get update before trying to get podman.
Make sure ~/.config dir exists.
Don't configure our selinux policy changes if selinux is absent or not enforcing.
Set PACKER_DIR variable so that validate-runenv-image-packer works
Set docker-init location based on linux distribution type.
Use 2 invocations of sg to git use into the docker group and then return us to our normal primary group.
Add validate runenv step to the km-ci-workflow

With these changes the payloads directory validate-runenv-image and validate-runenv-image-withpacker targets work.

This PR is the result of piecing branch paulp/validate-images-with-podman and branch serge/validate-runenv-2 together.


paulpopelka:
> `?` instead of `*` would be more precise. Otherwise things like `/dev/k_whatever_m` would match too. No big deal either way as this is just a message in the log though

Yes.  That occurred to me later.  This was put in to debug a failure claiming the device file didn't exist.  
Once I put the ls command in there the test passed and had been passing.
I assume this is some kind of intermittent problem.

gnode1 (approved):

paulpopelka:
Fixes #1124

PR #1334, opened 2021-08-30
use --runtime=krun in docs, scripts and such

gnode1:
Looked for `--device=/dev/kvm` and such everywhere, and change to `--runtime=krun` where appropriate. 

One noted area where is is *not* appropriate is our tests directory and corresponding test targets. In these cases we assume we are testing newly build km and we call in explicitly with the specific arguments.

In demos and such we should use the runtime.


paulpopelka:
Can you describe what you did to test this change?

sv641 (approved):

gnode1:
> Can you describe what you did to test this change?

The testing was done manually going through demo scenarios for java spring boot snapshot, tensorflow inference app, and such. Hopefully changes in the docs cover the manual testing steps

PR #1336, opened 2021-09-01
Change the bats sisgsuspend test to retry if there is a race related failure

paulpopelka:
Also added a message if virtual machine creation fails with EBUSY.
EBUSY probably results when another virtual machine manager is running.

### DIFF
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -603,6 +603,9 @@ void km_machine_setup(km_machine_init_params_t* params)
       km_errx(1, "KVM: API version mismatch");
    }
    if ((machine.mach_fd = km_internal_fd_ioctl(machine.kvm_fd, KVM_CREATE_VM, NULL)) < 0) {
+      if (errno == EBUSY) {
+         km_warnx("Is another virtual machine manager currently in use?");
+      }
       km_err(1, "KVM: create VM failed");
    }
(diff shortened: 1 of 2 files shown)


km/km_cpu_init.c
```diff
@@ -603,7 +603,7 @@ void km_machine_setup(km_machine_init_params_t* params)
       km_errx(1, "KVM: API version mismatch");
    }
    if ((machine.mach_fd = km_internal_fd_ioctl(machine.kvm_fd, KVM_CREATE_VM, NULL)) < 0) {
-      km_err(1, "KVM: create VM failed");
+      km_err(1, "KVM: create VM failed.%s", errno == EBUSY ? " Is another virtual machine manager currently in use?" : "");
```
gnode1, line 606: I don't understand the message here. We just opened the `/dev/kvm` and got`.mach_fd`. Any process can open it and create a VM, AFAIK.

gnode1 (dismissed):
With a question

km/km_cpu_init.c
paulpopelka: I've found that if a vagrant vm is running and then I try to run a payload with km this ioctl fails with EBUSY.
At first I didn't understand what was wrong.
Then I though this is because vagrant is running a vm.
I changed the message to help people understand what might be wrong.

Should km and vagrant should coexist on the same machine?

gnode1 (approved):

PR #1339, opened 2021-09-03
more parallel CI jobs

gnode1:
Make packer driven CI steps run in parallel.
Add log gathering if something fails - not really tested as nothing failed after adding these.
Disable KKM packer driven run for now, until KKM setup on Fedora (selinux related) is complete - see `km-krun-kkm-validate-runenv` step in `km-ci-workflow.yaml`

### DIFF
--- a/tests/packer/km-az-test.pkr.hcl
+++ b/tests/packer/km-az-test.pkr.hcl
@@ -44,6 +44,11 @@ variable "timeout" {
   type        = string
   description = "Timeout for tests. Should be less that outer timeout, so packer cleans up resources"
 }
+variable "step" {
+  type        = string
+  description = "Step ID in action workflow, used to make name unique"
+  default     = ""
+}
 
 variables {
   ssh_user = "ubuntu"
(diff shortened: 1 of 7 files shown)


tests/packer/scripts/gather-logs.sh
```diff
@@ -13,23 +14,19 @@
...
-LOG ?= 0
-PACKER_BUILD = PACKER_LOG=${LOG} packer build
-
-test: .check_packer .check_image_version ## Test KKM build on AWS using Hashicorp Packer
-	${PACKER_BUILD} \
-		--var src_branch=${SRC_BRANCH} \
-		--var image_version=${IMAGE_VERSION} \
-		km-aws-test.pkr.hcl
+# Log environment
+env
```
paulpopelka, line 23: Could we add "uname -a"?

tests/packer/km-az-test.pkr.hcl
```diff
@@ -119,9 +124,9 @@ build {
   }
 
   provisioner "shell" {
-    script          = "packer/scripts/km-test.sh"
+    script = "packer/scripts/km-test.sh"
     // double sg invocation to get docker into the process's grouplist but not the primary group.
-    execute_command = "chmod +x {{ .Path }}; {{ .Vars }} sg docker -c 'sg ubuntu {{ .Path }}'"
+    execute_command = "chmod +x {{ .Path }}; {{ .Vars }} sg docker -c 'sg ${var.ssh_user} {{ .Path }}'"
```
paulpopelka, line 129: Just a note here.
The argument to sg is a groupname and var.ssh_user is a username.
This works because the username is the same as the user's primary groupname.
I guess this is some sort of convention in linux these days.
We could probably do something that is the equivalent of:

sg `id -n -g` xxxxxx

We would need to fiddle around with where we run id -n -g to ensure we didn't get "docker".

I'm not asking you to do anything here.  Just commenting about it.

paulpopelka (approved):

PR #1343, opened 2021-09-07
Add a shell function that retries apt-get until the lock file is available

paulpopelka:
apt-get doesn't retry if another instance is currently running so we have an apt-get wrapper function that retries.

### DIFF
--- a/container-runtime/podman_config.sh
+++ b/container-runtime/podman_config.sh
@@ -108,22 +108,20 @@ function selinux_is_enabled()
    return 1
 }
 
-# A small function to keep trying apt-get until othe apt-get'ers finish.
+# A small function to keep trying apt-get until other apt-get'ers finish.
 function persistent-apt-get
 {
-   MESSAGE="E: Could not get lock /var/lib/dpkg/lock-frontend."
    eflag=0
    if echo $- | grep -q "e"; then eflag=1; fi
-   echo eflag $eflag
 
-   rv=0
+   MESSAGE="E: Could not get lock /var/lib/dpkg/lock-frontend."
    APT_GET_OUTPUT=/tmp/apt-get-out-$$
    while true; do
       cp /dev/null $APT_GET_OUTPUT
-      if [ $eflag -ne 0 ]; then set +e; fi
+      [ $eflag -ne 0 ] && set +e
       apt-get $* >$APT_GET_OUTPUT 2>&1
       rv=$?
-      if [ $eflag -ne 0 ]; then set -e; fi
+      [ $eflag -ne 0 ] && set -e
       if test $rv -ne 0; then
          if grep -q "$MESSAGE" $APT_GET_OUTPUT; then
             # Another apt-get is running, try again.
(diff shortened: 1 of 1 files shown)


container-runtime/podman_config.sh
```diff
@@ -108,6 +108,36 @@ function selinux_is_enabled()
    return 1
 }
 
+# A small function to keep trying apt-get until othe apt-get'ers finish.
+function persistent-apt-get
+{
+   MESSAGE="E: Could not get lock /var/lib/dpkg/lock-frontend."
+   eflag=0
+   if echo $- | grep -q "e"; then eflag=1; fi
+   echo eflag $eflag
```
gnode1, line 117: debug lefover?

container-runtime/podman_config.sh
```diff
@@ -108,6 +108,36 @@ function selinux_is_enabled()
...
+
+   rv=0
+   APT_GET_OUTPUT=/tmp/apt-get-out-$$
+   while true; do
+      cp /dev/null $APT_GET_OUTPUT
+      if [ $eflag -ne 0 ]; then set +e; fi
```
gnode1, line 123: `[ $eflag -ne 0 ] && set +e`

container-runtime/podman_config.sh
```diff
@@ -108,6 +108,36 @@ function selinux_is_enabled()
...
+      apt-get $* >$APT_GET_OUTPUT 2>&1
+      rv=$?
+      if [ $eflag -ne 0 ]; then set -e; fi
```
gnode1, line 126: `[ $eflag -ne 0 ] && set -e`

gnode1 (approved):
LGTM with a very small nit - I think some of the checks would look better in `[ condition ] && action` form rather then `if then else`. But it is real minor

container-runtime/podman_config.sh
paulpopelka: Yes.  Will remove.

container-runtime/podman_config.sh
paulpopelka: Change is coming.

container-runtime/podman_config.sh
```diff
@@ -108,6 +108,34 @@ function selinux_is_enabled()
...
+# A small function to keep trying apt-get until other apt-get'ers finish.
+function persistent-apt-get
+{
+   eflag=0
+   if echo $- | grep -q "e"; then eflag=1; fi
+
+   MESSAGE="E: Could not get lock /var/lib/dpkg/lock-frontend."
+   APT_GET_OUTPUT=/tmp/apt-get-out-$$
+   while true; do
+      cp /dev/null $APT_GET_OUTPUT
```
sv641, line 120: I can always guarantee, I learn something new every time I look at review.

touch $APT_GET_OUTPUT
is more commonly used.

gnode1: touch wouldn't erase the content of the file. Here the first time touch would so, but in the loop we need to make sure file is empty

paulpopelka: Yes, but I want to make sure the file starts out empty for each apt-get run.  Yes I could remove the file and recreate too.

gnode1 (approved):

PR #1345, opened 2021-09-08
Don't verify a containers.conf file that we created from scratch.

paulpopelka:
podman_config.sh will create ~/.config/containers/containers.conf from scratch if it
doesn't exist.  Then the code would drop down and verify that needed contents were
there and print messages.  If we create the file from scratch don't issue messages saying
configuration is present.  It confuses people.


gnode1 (approved):
LGTM assuming tests pass

paulpopelka:
The tests failed because of an intermittent failure of the bats perf test.
Will restart the test.

PR #1361, opened 2021-10-13
DOC-ONLY-fix for Docker run of kontain examples

smijar:
There was a discrepancy in that the 'docker run' example would not instantiate the example programs because the language interpreter name was omitted from the command.  So either we should put the language interpreter in the base image CMD and receive args, or we have to specify it in the docs.  This patch specifies it in the docs, so folks can try out Kontain successfully.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -576,14 +576,14 @@ These examples demonstrate how to run a simple Python application using a pre-bu
 Check what `os.uname()` reports:
 
 ```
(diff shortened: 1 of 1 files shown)


gnode1:
Good catch! But I have a feeling we are better off changing the base image to specify the entry point. Let's discuss.

gnode1:
OTOH it looks like standard python container behaves the same way as ours, so never mind the previous comment

gnode1 (approved):

PR #1366, opened 2021-10-18
DOC ONLY - mention of fedora 34 using cgroupsv2

smijar:
have added comments on not having to configure cgroups v1 for fedora 34 as the moby-engine installed using dnf on fedora34 uses cgroups v2.  Using cgroups v1 could lead to errors on fedora 34.
Configuring fedora31/32/33 for using cgroups v1 for moby-engine is critical but not for fedora34.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -55,17 +55,18 @@ Install
 sudo dnf install -y moby-engine
 ```
 
-There are a bunch of glitched between latest docker and latest fedora 32 config.
+*** Additional configuration if you are using fedora31 or fedora32 or fedora33 (moby-engine on fedora34 uses cgroups v2) ***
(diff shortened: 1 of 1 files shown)


docs/build.md
```diff
@@ -55,17 +55,18 @@ Install
 sudo dnf install -y moby-engine
 ```
 
-There are a bunch of glitched between latest docker and latest fedora 32 config.
+*** Additional configuration if you are using fedora31 or fedora32 or fedora33 (moby-engine on fedora34 uses cgroups v2) ***
```
gnode1, line 58: I thought 31 was not affected. Are you sure?

gnode1 (approved):

PR #1375, opened 2021-11-08
Install KKM from a Kubernetes DaemonSet.

johnmuth81:
Install KKM as a k8s DaemonSet. This does a build from source and load from an init container in a DaemonSet. The pattern was taken from a Google cloud example (https://github.com/GoogleCloudPlatform/solutions-gke-init-daemonsets-tutorial).

This builds and installs on GKE `ubuntu_containerd` type node.

### DIFF
--- a/cloud/k8s/deploy/README.md
+++ b/cloud/k8s/deploy/README.md
@@ -3,8 +3,23 @@
 This directory tools to install the Kontain runtime on Kubernets clusters. The of installation are
 per-node DeamonSet objects. There are two DaemonSets defined here:
 
-- `k8s-deploy.yaml` - Install the Kontain runtime
 - `kkm-install.yaml` - Install the KKM driver
+- `k8s-deploy.yaml` - Install the Kontain runtime
+
+## Installing KKM
+
+KKM is a software-only virutalization driver that the Kontain runtime can use when hardware virtualization (KVM)
+is not availale or is undesirable. KKM is not a full, general purpose virtualization driver. KKM only supports
+the facilities needed by the Kontain runtime. 
+
+To install KKM on all nodes in kubernetes cluster, run:
+
+```
+$ kubectl apply -f cm-kkm-install.yaml
+$ kubektl apply -f km-install.yaml
+```
+
+## Installing Kontain Runtime
 
 * THIS IS A WORK IN PROGRESS *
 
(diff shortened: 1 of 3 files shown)


cloud/k8s/deploy/README.md
```diff
@@ -1,4 +1,27 @@
...
+
+## Installing KKM
+
+KKM is a software-only virutalization driver that the Kontain runtime can use when hardware virtualization (KVM)
+is not availale or is undesirable. KKM is not a full, general purpose virtualization driver. KKM only supports
+the facilities needed by the Kontain runtime. 
+
+To install KKM on all nodes in kubernetes cluster, run:
+
+```
```
gnode1, line 17:
```bash

File: cloud/k8s/deploy/cm-kkm-install.yaml [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+    #!/usr/bin/env bash
+
+    set -x
+    set -euo pipefail
+
+    ROOT_MOUNT_DIR="${ROOT_MOUNT_DIR:-/root}"
+
+    echo "Installing dependencies"
+    apt-get update
+    apt-get install -y make gcc git
```
gnode1, line 34: What we discovered in our CI is that these occasionally fail when there is "unattended upgrade" running. In CI we implemented error check and loop. I don't know it this really applies here, let's discuss.

gnode1 (commented):
Is this ready to go, or still WIP? The TODO: clause makes me wonder.

cloud/k8s/deploy/cm-kkm-install.yaml
smijar: Something to think about: Not sure if an update to the OS is a good idea inside a Kubernetes deployment manifest.  Nodes probably needs to be upgraded or updated on a different schedule outside the purview of Kubernetes.

johnmuth81: This isn't an upgrade. All `apt-get update` does is update the package indexes so `apt-get install` is pulling from fresh places. `apt-get upgrade` is the command that upgrades the software on the local box. We aren't doing that.

johnmuth81:
> Is this ready to go, or still WIP? The TODO: clause makes me wonder.

This is ready to go. What needs changing is the km installation. It turns out that the kata-container install script that I based ours on assumes that pods can access the k8s control plane. This is a bad idea from a security POV and it disallowed by default by GKE and k3s. I'll have another PR this week addressing that issue.

smijar:
> > Is this ready to go, or still WIP? The TODO: clause makes me wonder.
> 
> This is ready to go. What needs changing is the km installation. It turns out that the kata-container install script that I based ours on assumes that pods can access the k8s control plane. This is a bad idea from a security POV and it disallowed by default by GKE and k3s. I'll have another PR this week addressing that issue.

Yup, I had seen that when trying to run it in K3s, that it was trying to run "kubectl get nodes" and failing.

gnode1 (approved):

johnmuth81:
This needs to be redone re-done to handle binary release bits.

PR #1376, opened 2021-11-08
prepare separate artifact for binary only  installs

gnode1:


johnmuth81 (approved):
lgtm

johnmuth81 (approved):
lgtm. thanks for doing this!

Issue #1374, opened 2021-11-05
Add udev to k8s daemonset 


johnmuth81:
echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666"' > /tmp/rules
sudo mv /tmp/rules /etc/udev/rules.d/99-perm.rules
sudo udevadm control --reload-rules && sudo udevadm trigger

PR #1378, opened 2021-11-09
Add udev config for KVM to k8s install DaemonSet

johnmuth81:
Fixes: #1374

### DIFF
--- a/cloud/k8s/deploy/k8s-deploy.sh
+++ b/cloud/k8s/deploy/k8s-deploy.sh
@@ -67,8 +67,12 @@ function configure_cri_runtime() {
 		configure_containerd
 		;;
 	esac
-	[ -c /host-devices/kvm ] && chmod 666 /host-devices/kvm
-	[ -c /host-devices/kkm ] && chmod 666 /host-devices/kkm
+
+        [ -c /host-dev/kvm ] && chmod 666 /host-dev/kvm
+        mkdir -p /etc/udev/rules.d
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666"' | tee /etc/udev/rules.d/99-perm.rules
+        udevadm control --reload-rules && udevadm trigger
+
 	systemctl daemon-reload
 	systemctl restart "$1"
 }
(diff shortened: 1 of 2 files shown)


sv641 (approved):

cloud/k8s/deploy/k8s-deploy.sh
```diff
@@ -67,8 +67,12 @@ function configure_cri_runtime() {
 		configure_containerd
 		;;
 	esac
-	[ -c /host-devices/kvm ] && chmod 666 /host-devices/kvm
-	[ -c /host-devices/kkm ] && chmod 666 /host-devices/kkm
+
+        [ -c /host-dev/kvm ] && chmod 666 /host-dev/kvm
```
gnode1, line 71: This should be superfluous, the `udevadm trigger` should trigger the appropriate event and udev itself should change the permissions.

cloud/k8s/deploy/k8s-deploy.sh
```diff
@@ -67,8 +67,12 @@ function configure_cri_runtime() {
...
+        mkdir -p /etc/udev/rules.d
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666"' | tee /etc/udev/rules.d/99-perm.rules
+        udevadm control --reload-rules && udevadm trigger
```
gnode1, line 74: This runs in the container, I wonder if does what we need to happen. IOW which udev does it kick?

johnmuth81: The YAMP file maps the node's `/var/run/dbus` and `/run/system` into the DaemonSet container. The reason for this is so things like `systemctl` and `udevadm` run from the DamonSet end up talking to the Node's daemons.

cloud/k8s/deploy/k8s-deploy.sh
johnmuth81: What about if udev isn't running at all? Isn't that possible?

cloud/k8s/deploy/k8s-deploy.sh
gnode1: `udevadm control --reload-rules` will reload this container's udev, not the host's, which means the host's one will not know of the change

johnmuth81: Latest commit maps udev socket so udevadm talks to Node daemon.

gnode1 (approved):

PR #1383, opened 2021-11-11
recover tensorflow snapshot demo

gnode1:
This is stop gap to have the tensorflow inference app with snapshot demo going.

The issues is that the TF code we have works with older TF (2.4.x) (current is 2.7), and python 3.8 (current is 3.9).

To do it right we need to port the tensorflow app to current tensorflow version. Then compile our version of that tensorflow for python 3.9.


johnmuth81 (approved):
lgtm.

Go on vacation already!

PR #1402, opened 2021-11-15
WIP: KKM install using kkm.run packaged in k8s DaemonSet

johnmuth81:
Includes CI test using minikube.


johnmuth81:
dump this

johnmuth81 (commented):
Re-open to try

johnmuth81:
It doesn't look like the DaemonSet approach to kkm install is going to work in the general case, although it may work in specific cases. Close this to re-think.

PR #1404, opened 2021-11-15
Install KKM with kkm.run using a k8s DaemonSet

### LINKED ISSUE #1386: KKM Install DamonSet for Kubernetes
Create a DaemonSet that compiles and installs KKM from a `kkm.run` file. This DeamonSet will be used as part of the Kontain installation procedure for Kubernetes when software virtualization is called for.

johnmuth81:
Fixes #1386


sv641 (approved):

johnmuth81:
It doesn't look like a general damonset solution to KKM install will work. Close this and re-address.

PR #1414, opened 2021-11-29
clone3() and changes in sigaltstack to work with glibc 2.34

gnode1:
Fedora 35 brings in glibc-2.34, which uses clone3() and dynamic stack size. This makes changes to support theses.

Tested manually running tests on Fedora 35. Separate clone3() test isn't implemented s musl has no wrapper for it.

There are still some other failures on Fedora 35 static glibc tests so this is not fully done yet.

### DIFF
--- a/km/km.h
+++ b/km/km.h
@@ -31,6 +31,7 @@
 #include <sys/param.h>
 #include <sys/types.h>
 #include <linux/kvm.h>
+#include <linux/sched.h>
 
 #include "bsd_queue.h"
 #include "km_elf.h"
@@ -385,6 +386,7 @@ int km_clone(km_vcpu_t* vcpu,
              km_gva_t ptid,
              km_gva_t ctid,
              unsigned long newtls);
+long km_clone3(km_vcpu_t* vcpu, struct clone_args* cl_args);
 uint64_t km_set_tid_address(km_vcpu_t* vcpu, km_gva_t tidptr);
 void km_exit(km_vcpu_t* vcpu);
 
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -505,6 +505,9 @@ int km_vcpu_clone_to_run(km_vcpu_t* vcpu, km_vcpu_t* new_vcpu)
    new_vcpu->regs = vcpu->regs;
    new_vcpu->regs.rsp = sp;
    *((uint64_t*)km_gva_to_kma_nocheck(sp)) = 0;   // hc return value
+   // syscall trap uses this to restore RDX 
+   *((uint64_t*)km_gva_to_kma_nocheck(sp + 0x18)) =
+       *((uint64_t*)km_gva_to_kma_nocheck(vcpu->regs.rsp + 0x18));
    km_vmdriver_clone(vcpu, new_vcpu);
    new_vcpu->regs_valid = 1;
 
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -40,6 +40,7 @@
 
 // Set CPUID VendorId to this. 12 chars max (sans \0) to fit in 3 register: ebx,ecx,edx
 static const char cpu_vendor_id[3 * sizeof(u_int32_t) + 1] = "Kontain";
+int set_cpu_vendor_id;
 
 km_machine_t machine = {
     .kvm_fd = -1,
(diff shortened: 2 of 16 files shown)


payloads/java/Makefile
```diff
@@ -82,8 +82,8 @@ fromsrc: .check ${JDK_VERSION} jtreg ${KM_BIN} ${KM_LDSO}
 	fi
 	@# -fcommon gcc flag is needed for gcc10, see https://bugs.openjdk.java.net/browse/JDK-8235903
 	cd ${JDK_VERSION} && bash configure \
-		--disable-warnings-as-errors --with-native-debug-symbols=none \
-		--with-extra-cflags='-fcommon'  \
+		--disable-warnings-as-errors --with-native-debug-symbols=internal \
```
johnmuth81, line 85: Yeah, I see what you mean. I think this was just source code entropy in action. Good catch.

johnmuth81 (commented):
I think the java thing was just code drift. Good catch.

Added some other comments.

km/km.h
```diff
@@ -31,6 +31,7 @@
 #include <sys/param.h>
 #include <sys/types.h>
 #include <linux/kvm.h>
+#include <linux/sched.h>
```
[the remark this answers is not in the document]
gnode1: right

km/km_cpu_init.c
```diff
@@ -667,10 +670,12 @@ void km_machine_setup(km_machine_init_params_t* params)
             __get_cpuid(entry->function, &entry->eax, &entry->ebx, &entry->ecx, &entry->edx);
             break;
          case 0x0:
+            __get_cpuid(entry->function, &entry->eax, &entry->ebx, &entry->ecx, &entry->edx);
+
             km_infox(KM_TRACE_KVM, "Setting VendorId to '%s'", cpu_vendor_id);
```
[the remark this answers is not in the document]
gnode1: I'm still working on this one.

gnode1:
I think this is ready to be merged. Our tests work fine. The node tests, even native node tests, require changes in the system SSL settings, but even then node v12 still has one test that fails. I'm skipping this test or now. I think the right answer there is to move to more current version of node, in a separate PR

km/km_cpu_init.c
```diff
@@ -667,10 +671,12 @@ void km_machine_setup(km_machine_init_params_t* params)
...
-            km_infox(KM_TRACE_KVM, "Setting VendorId to '%s'", cpu_vendor_id);
-            memcpy(&entry->ebx, cpu_vendor_id, sizeof(u_int32_t));
-            memcpy(&entry->edx, cpu_vendor_id + sizeof(u_int32_t), sizeof(u_int32_t));
-            memcpy(&entry->ecx, cpu_vendor_id + 2 * sizeof(u_int32_t), sizeof(u_int32_t));
+            if (set_cpu_vendor_id != 0) {
+               km_infox(KM_TRACE_KVM, "Setting VendorId to '%s'", cpu_vendor_id);
+               memcpy(&entry->ebx, cpu_vendor_id, sizeof(u_int32_t));
+               memcpy(&entry->edx, cpu_vendor_id + sizeof(u_int32_t), sizeof(u_int32_t));
+               memcpy(&entry->ecx, cpu_vendor_id + 2 * sizeof(u_int32_t), sizeof(u_int32_t));
+            }
```
sv641, line 679: how does it work when set_cpu_vendor_id is not set.

km/km_hcalls.c
```diff
@@ -1432,6 +1432,23 @@ static km_hc_ret_t clone_hcall(void* vcpu, int hc, km_hc_args_t* arg)
    return HC_CONTINUE;
 }
 
+static km_hc_ret_t clone3_hcall(void* vcpu, int hc, km_hc_args_t* arg)
+{
+   // long syscall(SYS_clone3, struct clone_args *cl_args, size_t size);
+   struct clone_args* cl_args = km_gva_to_kma(arg->arg1);
+
+   if (cl_args == NULL) {
```
sv641, line 1440: we generally do end of structure check to be in bounds.

km/km_cpu_init.c
gnode1: `entry` is an element from `machine.entries` array which is set by `KVM_GET_SUPPORTED_CPUID` ioctl a little above, so it is already set up. I'm not sure why 0x80000002, 0x80000003 and 0x80000004 are treated special, haven't really dug into it

km/km_hcalls.c
gnode1: Not sure this is actually correct. I see a lot of places where we only check the address. Can you point at examples?

sv641: uname hcall.

sv641 (approved):

Issue #1411, opened 2021-11-24
statically link KRUN


johnmuth81:
KRUN is dynamically linked. This presents installation problems with shared libraries. It would be better if KRUN were statically linked.

PR #1418, opened 2021-12-03
KM Binary Tar for k8s with strictly static binaries

johnmuth81:
Fixes: #1411 

Rework binary packaging and CI to integrate statically linked KRUN into KM source base.

This actually got a bit more involved than I thought at first. Here are the major changes:

* Update kontainapp/crun level to 5a75bcc29962f15336d409adee32c2072b704784
* Get rid of container-runtime test-remote.py. This was trying to build and test kontainapp/crun on an Azure VM (even though it was already built and tested on github runner). The compile was failing with old headers. Rather than tracing it down it down, I removed it.
* Build static KRUN in parallel with the existing KM build/test. This makes sense since the `nix` based KRUN static build takes a long time. After both the km and static krun are built, a binary release tar is synthesized that contains all static binaries.
* A text file that contains the output of `git submodule status` is included  on the new binary release tar file. This is sufficient to pin the version of KKM required for KM and is much smaller than kkm.run.


.github/workflows/km-ci-workflow.yaml
```diff
@@ -161,10 +150,95 @@ jobs:
...
+      - uses: actions/cache@v2
+        with:
+          path: .cache
+          key: nix-v1-${{ hashFiles('container-runtime/crun/nix/nixpkgs.json') }}
+
+      - run: |
+          set -x
+          sudo find / -name crun -type f
+          CRUN_DIR=$(pwd)/container-runtime/crun
+          if [[ -z $(ls -A /nix) ]]; then sudo docker run --rm --privileged -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix; fi
```
gnode1, line 183: What;s the point of the `if` here? I assume the environment is the same every time, the directory is either there or not.

.github/workflows/km-ci-workflow.yaml
```diff
@@ -161,10 +150,95 @@ jobs:
...
+          sudo docker run --rm --privileged -v /nix:/nix -v ${CRUN_DIR}:${CRUN_DIR} -w ${CRUN_DIR} nixos/nix nix build --print-build-logs --file nix/ --arg enableSystemd false
+          find . -name result
+          find $(find . -name result)
```
gnode1, line 186: What's the reason for the two finds?

.github/workflows/km-ci-workflow.yaml
```diff
@@ -136,17 +136,6 @@ jobs:
...
-          make kkm-pkg
-          make release
-          mkdir /tmp/release
-          cp -r build/kontain.tar.gz /tmp/release
-          mkdir /tmp/bin_release
-          cp -r build/kontain_bin.tar.gz /tmp/bin_release
-
       - uses: actions/upload-artifact@v2
         with:
           name: km
```
gnode1, line 141: Three lines below - where is this krun coming from and what's the point?

.github/workflows/km-ci-workflow.yaml
```diff
@@ -456,6 +517,13 @@ jobs:
         kubectl apply -k cloud/k8s/deploy/kontain-deploy/ci
         sleep 20
 
+    # uncomment to get logs from damonset.
```
gnode1, line 520: Maybe `if env.SOMETHING == true` instead?

gnode1:
The sha mentioned in the description - is it simply the sha we synced with? And related question - when we build krun, where is the sha/tag/branch come from?

.github/workflows/km-ci-workflow.yaml
johnmuth81: This was lifted directly from the CRUN CI script. Perhaps a comment to that effect would be helpful.

.github/workflows/km-ci-workflow.yaml
```diff
@@ -161,10 +150,95 @@ jobs:
...
+          submodules: recursive
+
+      - uses: actions/cache@v2
+        with:
+          path: .cache
+          key: nix-v1-${{ hashFiles('container-runtime/crun/nix/nixpkgs.json') }}
+
+      - run: |
+          set -x
+          sudo find / -name crun -type f
```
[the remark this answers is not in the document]
johnmuth81: This is debugging I should have removed.

.github/workflows/km-ci-workflow.yaml
johnmuth81: More debugging I should remove.

johnmuth81:
> The sha mentioned in the description - is it simply the sha we synced with? And related question - when we build krun, where is the sha/tag/branch come from?

1. What ! mentioned in the comment is the output of `git submodule status`. eg.: 
```
 5a75bcc29962f15336d409adee32c2072b704784 container-runtime/crun (0.14.1-685-g5a75bcc)
 37730e5f349c8a12f3a64b2dbde9e8a337704b95 kkm (heads/master)
 b19da8e362acae8944c60a40cf5c40c8f5ebeb44 lib/mimalloc/mimalloc (v1.7.1-1-gb19da8e)
 4721cce4af0c4d118a96316d087f1ad30f1a6932 runtime/musl (remotes/origin/kontain)
 565f2863288331684cf5b03560b26151a09ea810 tests/bats (v1.1.0-438-g565f286)
 9f88b4207da750093baabc4e3f41bf68f0dd3630 tests/bats-assert (v0.2.0-5-g9f88b42)
 004e707638eedd62e0481e8cdc9223ad471f12ee tests/bats-support (v0.3.0-1-g004e707)
 bd7bc060d67078562625984490c209a7459bd822 tests/greatest (v1.5.0-1-gbd7bc06)
```
The important bit of information here is the the 2nd line. It tells us what version of KKM matches KM from a git commit SHA POV. So if we need to build KKM, we can do `git clone <blah>; cd <blah>; git checkout <kkm-sha>` and we know we'll build a version that matches (assuming the kernel headers exist). Srini and I talked about this a couple weeks ago.

.github/workflows/km-ci-workflow.yaml
johnmuth81: That's the dynamically linked krun that we were building before and we are still using for testing in the km-build step. I haven't transitioned that. If you think we need to we could open an issue.

.github/workflows/km-ci-workflow.yaml
johnmuth81: I'll look into that. In the meantime, I'll push another commit with the debugging stuff removed.

.github/workflows/km-ci-workflow.yaml
gnode1: OK. I'm thinking krun is better off being separate tree with build artifact, and this flow should just pull that and use whenever krun is used. I know you guys decided not to do that, not really sure why. 

Either way it is good as is for now.

gnode1 (approved):

PR #1424, opened 2021-12-09
relax {p}read/write check for zero buf

gnode1:
I turns out `write(fd, NULL, 0)` is supposed to succeed and return 0. So out check for the args is too zealous. And yes, node v16 does make such calls.

Regular tests. Also specific test with node v16 now pass. There are more issues there so node isn't ready yet, but I thought this is important on its own.

### DIFF
--- a/km/km_hcalls.c
+++ b/km/km_hcalls.c
@@ -147,7 +147,7 @@ static km_hc_ret_t prw_hcall(void* vcpu, int hc, km_hc_args_t* arg)
    // ssize_t pwrite(int fd, const void* buf, size_t count, off_t offset);
    // arg->hc_ret = __syscall_4(hc, arg->arg1, km_gva_to_kml(arg->arg2), arg->arg3, arg->arg4);
    void* buf = km_gva_to_kma(arg->arg2);
-   if (buf == NULL) {
+   if (buf == NULL && arg->arg3 != 0) {
       arg->hc_ret = -EFAULT;
       return HC_CONTINUE;
    }


sv641 (approved):

paulpopelka:
From the linux wite syscall man page:
```
If count is zero and fd refers to a regular file, then write() may return a failure status if one of the errors  below  is  detected.   If  no  errors are detected, or error detection is not performed, 0 will be returned without causing any other effect.  If count is zero and fd refers to a file other than a regular file, the results are not specified.
```
The above allows this behavior.

sv641:
Hopefully this will put this to rest.
Original att code.

rdwr(mode)
register mode;
{
       register struct file *fp;
       register struct inode *ip;
       register struct a {
               int     fdes;
               char    *cbuf;
               unsigned count;
       } *uap;
       int     type;

       uap = (struct a *)u.u_ap;
       fp = getf(uap->fdes);
       if (fp == NULL)
               return;
       if ((fp->f_flag&mode) == 0) {
               u.u_error = EBADF;
               return;
       }
       u.u_base = (caddr_t)uap->cbuf;
       u.u_count = uap->count;
       u.u_segflg = 0;
       u.u_fmode = fp->f_flag;
       ip = fp->f_inode;
       type = ip->i_mode&IFMT;
       if (type==IFREG || type==IFDIR) {
               plock(ip);
               if ((u.u_fmode&FAPPEND) && (mode == FWRITE))
                       fp->f_offset = ip->i_size;
       } else if (type == IFIFO) {
               plock(ip);
               fp->f_offset = 0;
       }
       u.u_offset = fp->f_offset;
       if (mode == FREAD)
               readi(ip);
       else
               writei(ip);
       if (type==IFREG || type==IFDIR || type==IFIFO)
               prele(ip);
       fp->f_offset += uap->count-u.u_count;
       u.u_rval1 = uap->count-u.u_count;
       u.u_ioch += (unsigned)u.u_rval1;
       if (mode == FREAD)
               sysinfo.readch += (unsigned)u.u_rval1;
       else
               sysinfo.writech += (unsigned)u.u_rval1;
}

writei has this loop

        case IFDIR:
       while (u.u_error==0 && u.u_count!=0) {


Behavior has been this for even before I was born.

sv641:
case IFBLK:
       case IFREG:
       case IFDIR:
       while (u.u_error==0 && u.u_count!=0) {

PR #1429, opened 2021-12-14
Start a toolkit for Kubernetes

johnmuth81:
First tool is a node file system examination tool.

### DIFF
--- a/cloud/k8s/tools/README.md
+++ b/cloud/k8s/tools/README.md
@@ -1,0 +1,11 @@
+# Tools for Kubernetes
+
+This directory contains tools to diagnose and debug kubernetes clusters.
+
+## hacknode.yaml
+
+This creates a privileged DaemonSet in the `kube-system` namespace with a base name of `kontain-node-hack`
+to run on all Nodes. The daemon set mounts the Node's root file system `/` as `/root` and sleeps forever.
+
+The pod(s) running the DamonSet are found with `kubectl get pod -A -o wide`. A shell can them be started
+on a DamonSet pod with `kubectl exec -it`.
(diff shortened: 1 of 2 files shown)


sv641 (approved):

cloud/k8s/tools/hack_node.yaml
```diff
@@ -0,0 +1,48 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kontain-node-hack
+  namespace: kube-system
```
smijar, line 21: Maybe use a different namespace for this like 'default' else the developer might forget to clean up the pod in kube-system thus affecting the quota for the namespace.

smijar: If you leave namespace blank it defaults to the current context.

cloud/k8s/tools/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This creates a privileged DaemonSet in the `kube-system` namespace with a base name of `kontain-node-hack`
+to run on all Nodes. The daemon set mounts the Node's root file system `/` as `/host-root` and sleeps forever.
+
+The pod(s) running the DamonSet are found with `kubectl get pod -A -o wide`. A shell can them be started
+on a DamonSet pod with `kubectl exec -it`.
+
+### Tips and Tricks
+
+- Once logged into the DaemonSet pod, the Nodes `/proc` is mounted as `/host-root/proc`, so the command `chroot /host-root ps -aef` will show all processes running on the Node.
```
smijar, line 15: Also found out that if the user uses $ chroot /host-root, then he can use the host's normal processes like a regular user.

johnmuth81: How so? Did `kill` work? (that would surprise me).

PR #1436, opened 2021-12-17
Fedora 35

gnode1:
Support Fedora 35 as host - native builds and tests. Main change is libstdc++ is again our own build, to make it compatible with new glibc.

Also of course withdocker.

Upped node to v16.13.1.


.github/workflows/km-ci-workflow.yaml
```diff
@@ -56,7 +56,7 @@ on:
         required: true
 
 env:
-  BUILDENV_IMAGE_VERSION: latest # use this for all buildenv containers
+  BUILDENV_IMAGE_VERSION: PR_1436 # use this for all buildenv containers
```
johnmuth81, line 59: Is this for debugging?

gnode1: This is to make this CI run use the new buildenv while maintaining the old one for the rest of the runs. It is documented in `image-targets.md`

sv641 (dismissed):

PR #1444, opened 2021-12-22
vfork

gnode1:
Turns our java uses vfork hcall for java.lang.Runtime.exec(). This makes vfork do the same thing as fork.

### DIFF
--- a/km/km_hcalls.c
+++ b/km/km_hcalls.c
@@ -2104,6 +2104,7 @@ const km_hcall_fn_t km_hcalls_table[KM_MAX_HCALL] = {
     [SYS_execve] = execve_hcall,
     [SYS_execveat] = execveat_hcall,
     [SYS_fork] = fork_hcall,
+    [SYS_vfork] = fork_hcall,
     [SYS_wait4] = wait4_hcall,
     [SYS_waitid] = waitid_hcall,
     [SYS_times] = times_hcall,


sv641 (approved):

km/km_hcalls.c
```diff
@@ -2104,6 +2104,7 @@ const km_hcall_fn_t km_hcalls_table[KM_MAX_HCALL] = {
     [SYS_execve] = execve_hcall,
     [SYS_execveat] = execveat_hcall,
     [SYS_fork] = fork_hcall,
+    [SYS_vfork] = fork_hcall,
```
paulpopelka, line 2107: I hate to say this.....  Are you going to add a vfork() test?
And then verify that vfork() does act differently from fork()?
I looked at the vfork() man page and the thing that stands out is the vfork()ing thread is suspended until the child process calls exec() or exits.

sv641: I hate to say this (; 
we are aliasing vfork to fork.(not implementing vfork semantics). What additional test would be helpful?

paulpopelka: You might have stated that we are intentionally not following vfork semantics in the PR description.
I guess that is just unreasonable.

gnode1: My understanding is that `vfork()` suspending the parent is an artifact of the early BSD implementation. They had to suspend the parent as they didn't want to create a child thread (not could they back then), so the child code was borrowing the parent thread all the way until `exec()`.

I don't think there is code out there relying in that. Instead AFAIK `vfork()` is simply "cheaper and faster" `fork()`. Other than that thread behavior `vfork()` requires certain behavior guarantees from the child. If we use regular `fork()` instead that still should be OK.

So, in short - I was hopping to get by without extra code for `vfork()` or extra tests, but I am open to change my mind if I'm overlooking something.

paulpopelka: Since we don't suspend the parent thread until the child exec's, the parent and child threads are running on the same stack at the same time potentially stomping on each other's locals, return addresses, whatever.
Which will probably lead to odd sporadic failures depending on which thread is scheduled or preempted as each thread moves ahead in its execution.

That is my concern.  I guess we can try this and see how it goes.

sv641: @paul, this is from man page. 

**The requirements put on vfork() by the standards are weaker than those put on fork(2), so an implementation where the two are  synonymous  is  compliant.**

sv641: The worst part of following man pages is the above statement also means you can implement **vfork semantics for fork**. But at the end of the day what you said is correct, does it work?

gnode1: @paulpopelka parent and child are not sharing anything - vfork aliased as fork so child runs in it's own stack and everything.

paulpopelka: I've been thinking about this and it dawned on me that vfork() won't have the problem I worried about.
As long as you don't depend on the oddities of vfork(), then it should be ok.

paulpopelka (approved):

PR #1448, opened 2021-12-30
Add shell-less dynamic base and jdk

gnode1:
The jdk runenv is now shell-less. It is used as base for log4j exploit demo, which makes the exploit fail one steps earlier - since there is no shell to execute.

To get the jdk with shell use jdk-shell runenv. This one is used for springboot demo.

### DIFF
--- a/payloads/dynamic-base-scratch/runenv.dockerfile
+++ b/payloads/dynamic-base-scratch/runenv.dockerfile
@@ -13,9 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-ARG RUNENV_IMAGE_VERSION=latest
 
-FROM kontainapp/runenv-jdk-11.0.8:${RUNENV_IMAGE_VERSION}
-COPY scripts /scripts
-EXPOSE 8080
-CMD ["java", "-cp", "/scripts", "SimpleHttpServer"]
+FROM scratch
+ENV KM_DO_SHELL NO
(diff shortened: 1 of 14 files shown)


johnmuth81 (approved):

payloads/dynamic-base-scratch/runenv.dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+FROM scratch
+ENV KM_DO_SHELL NO
+ADD --chown=0:0 libs.tar /opt/kontain/
+ADD empty_tmp /tmp/
```
smijar, line 20: wonder if there's a way to avoid this step.  Looks weird in a Dockerfile.

smijar (commented):
Noted that the JDK for log4j and spring boot demo is changed to kontainapp/runenv-jdk-shell-11.0.8:latest

payloads/dynamic-base-scratch/runenv.dockerfile
gnode1: This is our base Dockerfile, not really visible to users. We do need to make sure `/tmp/` is there. Many things don't work without it.

In theory we could make the tar image with it and simply ADD that. Let me know how strongly you feel about it.

PR #1459, opened 2022-01-13
enable SIGINT forwarding and cleanup

### LINKED ISSUE #1456: KM: Cannot exit docker container running in foreground with Ctrl-C like regular docker
**Describe the bug**
When running a regular docker container in the foreground, you can quit using Ctrl-C.  With a Kontain Container running in the foreground, you cannot quit it using the Ctrl-C hot key.  You have to start another shell window and use docker stop command.

gnode1:
regular tests, also confirmed manually ^C now works, so this resolves #1456


sv641 (approved):

smijar (approved):

johnmuth81 (approved):
lgtm

PR #1471, opened 2022-01-20
Fix concurrency errors in km_filesys around inuse and next internal fd

### LINKED ISSUE #1469: KM : inconsistent usage of km_file:inuse variable
km_file:inuse variable is used as a normal variable at times and at times it is used as a atomic variable.

we need to make code consistent.

It is also preferable to use atomic interface from stdatomic.h instead of __atomic interface.

johnmuth81:
Fixes #1469

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -76,6 +76,19 @@ static int km_fs_g2h_filename(const char* name, char* buf, size_t bufsz, km_file
 static int km_fs_g2h_readlink(const char* name, char* buf, size_t bufsz);
 
 /*
+ * Tells whether a file is inuse or not.
+ */
+int km_is_file_used(km_file_t *file)
+{
+   return (__atomic_load_n(&file->inuse, __ATOMIC_SEQ_CST) != 0);
+}
+
+void km_set_file_used(km_file_t *file, int val)
+{
+   __atomic_store_n(&file->inuse, val, __ATOMIC_SEQ_CST);
+}
+
+/*
  * Get file name for file descriptors that are not files - socket, pipe, and such
  */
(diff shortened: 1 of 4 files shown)


paulpopelka:
This fixes 1468 too.

sv641 (approved):

sv641 (commented):
why do we use __atomic instead of atomic.

we are using __ATOMIC_SEQ_CST.

km/km_filesys.c
```diff
@@ -1131,7 +1131,7 @@ uint64_t km_fs_dup3(km_vcpu_t* vcpu, int fd, int newfd, int flags)
    assert(name != NULL);
    ret = __syscall_3(SYS_dup3, host_fd, newfd, flags);
    if (ret >= 0) {
-      if (km_fs()->guest_files[ret].inuse != 0) {
+      if (km_is_file_used(&km_fs()->guest_files[ret]) != 0) {
          del_guest_fd(vcpu, ret);
       }
       ret = km_add_guest_fd(vcpu, ret, name, flags, ops);
```
paulpopelka, line 1137: It feels like this entire sequence: dup3(), is_file_used test, del_guest_fd(), add_guest_fd() should be an atomic operation.

If you have 2 (or more) threads dup3()ing onto the same newfd, the results would vary depending on whether the involved threads are preempted or not.  We would probably see assert failures in del_guest_fd() or km_add_guest_fd_internal().

johnmuth81: Yes this is totally valid, but here is what I'm going to do:

Going to merge without dup3 changes suggested here. Opened Issue #1479 to capture this.

PR #1475, opened 2022-01-25
Change km to use crun 1.4

paulpopelka:
Change km to use crun 1.4 which also used image-spec v1.0.2.
I thought this would eliminate the complaints from github about km still using v1.0.1.
Come to find out that tools/faktory's go.mod file still refers to image-spec v1.0.1
That is fixed with another PR.

Tested using crun's test suite.


paulpopelka:
I moved up to crun 1.4.2 to fix a problem with unexpect cgroup CPUshares value in crun.

paulpopelka:
Closing this PR since it is of no value.

PR #1490, opened 2022-01-28
switch to ubuntu 20.04

gnode1:
Ubuntu 2010 is deprecated. Our release box for ubuntu didn't build for v0.9.4 because repositories are closed. We also use 2010 in some internal builds. I'm replacing it with 20.04 which is LTS. Also rebuilding the base image for tests.

The failure was hidden, hopefully it won't be any more

### DIFF
--- a/tools/hashicorp/Makefile
+++ b/tools/hashicorp/Makefile
@@ -22,7 +22,7 @@ include ${TOP}/make/actions.mk
 all: vm-images
 
 # Use these boxes as sources for our boxes builds. (`generic/$os' is the box)
-SRC_OS := ubuntu2010 fedora32
+SRC_OS := ubuntu2004 fedora32
 # Note: SRC_OS needs to be synced with os_name in *pkrvars.hcl files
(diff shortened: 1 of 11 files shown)


tools/hashicorp/Makefile
```diff
@@ -34,13 +34,13 @@ product: ${ARTIFACTS} ## Prepares KM packaging artifacts (e.g. km tarball and kk
 	echo $(MAKE) MAKEFLAGS=$(MAKEFLAGS) -C ${TOP} release
 
 vm-images: .check_packer .check_tools product ## loads source boxes and creates boxes with KM installed
-	@set -x ; for os in ${SRC_OS} ; \
+	@set -ex ; for os in ${SRC_OS} ; \
```
sv641, line 37: is this intentional.

sv641 (approved):

tools/hashicorp/Makefile
gnode1: propagates errors to the make - otherwise the errors are silent. That's why we didn't notice bix build failure

Issue #1481, opened 2022-01-27
KM Install needs to include 'amzn' in list of supported distros


smijar:
We can now support Amazon linux 2 kernel version > 5, so we need the base install to include 'amzn' in the list of supported distros that we can install to.

smijar:
Need to modify kontain-install.sh to add 'amzn' as shown below:

Line 27 and 28
...
source /etc/os-release
[ "$ID" != "fedora" -a "$ID" != "ubuntu" -a "$ID" != "amzn" ] && echo "Unsupported linux distribution: $ID" && exit 1
....

PR #1497, opened 2022-02-03
#1481- added amzn to list of supported distros we can install to

smijar:
added amzn to list of supported distros we can install to


johnmuth81 (approved):
lgtm

PR #1512, opened 2022-02-14
reconfigure udev to make /dev/kvm 0666

gnode1:


johnmuth81:
What's driving this one? I'm a bit nervous about the security implications. Is there a groups/CAPS sort of solution to what you are trying to solve?

johnmuth81 (approved):
We already talked about this.

PR #1515, opened 2022-02-16
Disable KM guests from using LD_PRELOAD

### LINKED ISSUE #1514: Disable LD_PRELOAD for dynamic loaders by setting AT_SECURE
There is a class of exploits that use LD_PRELOAD to hijack processes. For example: https://thehackernews.com/2021/12/new-payment-data-sealing-malware-hides.html.

Setting the AT_SECURE entry to 1 in the Auxiliary Vector will disable LD_PRELOAD in dynamic loaders. The Linux kernel uses AT_SECURE for suid programs. We should use it for everything.

johnmuth81:
The Linux kernel disables LD_PRELOAD for SUID programs since allowing it would be an easy way to escalate privileges. 
We decided that KM should disallow LD_PRELOAD as a matter of course.
 
Fixes #1514


johnmuth81:
Bzzzt!!! It turns out that JDK actually uses LD_PRELOAD, so we can't get rid of it....

PR #1517, opened 2022-02-17
DOC-ONLY-install KKM using kkm.run consistently

smijar:


gnode1 (commented):
Should we fix kkm.run to take these steps instead?

smijar:
That would be ideal yes, I think.  I did notice that kkm.run was downloading kernel-version-16-devel on a kernel version 10 system.

gnode1:
> Should we fix kkm.run to take these steps instead?

Yes, we should fix the install script so it will take care of that.

smijar:
So to do this we have to update [lines](https://github.com/kontainapp/kkm/blob/32970b6dbc889c8d32a2049ed6f9546d8b8bf62c/installer/build-script.sh#L55-L68) 

```shell
# for dnf we have to include: 
$ dnf update --exclude=kernel*

# for yum, as well:
$ yum update --exclude=kernel*

# for apt luckily it looks 'apt update' does not update kernel, but to be sure we can add this, but needs a little testing: 
$ apt-mark hold linux-image-generic linux-headers-generic
```

docs/kkm_install_amznlinux2_fedora
```diff
@@ -0,0 +1,31 @@
...
+yum -y --exclude=kernel* update
+
+# check if it needs rebooting while yum not running
+(test ! -f /var/run/yum.pid && needs-restarting -r) || (reboot)
+```
+
+For Fedora:
+```shell
+# install kernel-devel for current package
+sudo dnf install -y "kernel-devel-uname-r == $(uname -r)" dnf-utils
```
gnode1, line 22: `sudo` here is unnecessary - `dnf install` requires root, so does reboot.

docs/kkm_install_amznlinux2_fedora
```diff
@@ -0,0 +1,31 @@
...
+This prevents errors that pop up during the build with header locations when building KKM on target system.
+
+For Amazon Linux 2:
+```shell
+yum install -y "kernel-devel-uname-r == $(uname -r)" yum-utils
+# update without updating kernel (IMPORTANT FOR KKM)
+yum -y --exclude=kernel* update
+
+# check if it needs rebooting while yum not running
+(test ! -f /var/run/yum.pid && needs-restarting -r) || (reboot)
```
gnode1, line 16: ASAIK the parenthesis are not necessary. Also, we are not rebooting if yum is running. What happens then?

PR #1524, opened 2022-03-02
kustomize changes to create overlays for various gke, aks, eks, minikube, crio, k3s, kind

smijar:
Added support for gke, aks, eks, crio, k3s, minikube, kind, ci, by consolidating all usecases under base, k3s, kkm, km-crio overlays.

will need to modify ci to use overlays/ci


smijar:
Got it.  Will do that as part of this PR. Thanks.

smijar:
> You'll need to update `.github/workflows/km-ci-workflow.yaml` line 492 to account for moving where `runtime-class.yaml` lives. Should be a simple, one line fix. That's why the CI failed.

Done.  And the CI seems to pass now.

johnmuth81 (approved):
lgtm

PR #1534, opened 2022-03-15
km with GVA == KMA

gnode1:
This builds KM as a static PIE, which has the effect of loading in very high memory due to Linux ASLR.

This allows allocating guest memory in KM at address 0, so that physical guest address (PVA) is the same as in km. This is controlled by KM_GPA_AT_16T macro. If defined, km reverts to old physical address schema.

If KM_HIGH_GVA macro is not defined, the virtual guest addresses (GVA) are equal to PVA. If it is defined km reverts to old schema with high GVA up at 128T.

Which means no virtual address translation needed in KM.

Also cleaned up and (I think) simplified memory management code.

Note that the kkm code is not adapted to this yet, so many tests fail. Hence this PR is still WIP.

I'd appreciate a review to make sure I'm not doing something crazy.

### DIFF
--- a/km/km_guest_asmcode.s
+++ b/km/km_guest_asmcode.s
@@ -89,7 +89,7 @@ intr_hand CP, 21
 /*
  * Table used by KM to build IDT entries.
  */
-    .section .km_guest_data, "da", @progbits
+    .section .km_guest_data, "dwa", @progbits
     .align 16
     .type __km_interrupt_table, @object
     .global __km_interrupt_table
--- a/km/km_mem.c
+++ b/km/km_mem.c
@@ -49,8 +49,9 @@
  *   that GB
  *
  * When using kvm driver
- * - We are forced to stay within width of the CPU physical memory bus. We
- *   determine that the by analyzing CPUID and store in machine.guest_max_physmem
+ * - We are forced to stay within width of the CPU physical memory bus. We determine that by analyzing
+ * CPUID and store in machine.guest_max_physmem. Currently the code only supports up to 512GB.
+ *
  * When using kkm driver
  * - fixed 512GB of memory is supported.
  */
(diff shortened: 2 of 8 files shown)


km/km_guest_asmcode.s
```diff
@@ -89,7 +89,7 @@ intr_hand CP, 21
 /*
  * Table used by KM to build IDT entries.
  */
-    .section .km_guest_data, "da", @progbits
+    .section .km_guest_data, "dwa", @progbits
```
sv641, line 92: do you want to change permissions to write on trap table?

sv641 (approved):

km/km_guest_asmcode.s
johnmuth81: I'm not sure when this went in or why. @gnode1 ?

gnode1: static pie link was complaining otherwise

sv641 (dismissed):
Are we changing the lower memory to 256GB and upper memory to 256GB.
Previously we had combined max to  512GB.

sv641 (dismissed):

km/km_mem.c
```diff
@@ -49,8 +49,9 @@
  *   that GB
  *
  * When using kvm driver
- * - We are forced to stay within width of the CPU physical memory bus. We
- *   determine that the by analyzing CPUID and store in machine.guest_max_physmem
+ * - We are forced to stay within width of the CPU physical memory bus. We determine that the by
```
paulpopelka, line 52: Can you remove "the" from "that the by".  Doesn't feel right.

tests/km_core_tests.bats
```diff
@@ -584,9 +584,9 @@ fi
       --ex="target remote :$km_gdb_port" \
       --ex="source cmd_for_protected_mem_test.gdb" --ex=q
    assert_success
-   assert_line --partial "first word  0x7fffffbfc000:	0x1111111111111111"
-   assert_line --partial "spanning pages  0x7fffffbfcffc:	0xff0000ffff0000ff"
-   assert_line --partial "last word  0x7fffffbfdff8:	0xeeeeeeeeeeeeeeee"
+   assert_line --regexp "first word  0x7ffff?f?bfc000:	0x1111111111111111"
+   assert_line --regexp "spanning pages  0x7ffff?f?bfcffc:	0xff0000ffff0000ff"
+   assert_line --regexp "last word  0x7ffff?f?bfdff8:	0xeeeeeeeeeeeeeeee"
```
paulpopelka, line 589: Could you add a note to explain why the "?" in the assert_line's are there.
Whoever looks at this in a year is probably wonder why this is here.

km/km_mem.h
```diff
@@ -160,10 +182,10 @@ static inline km_gva_t gpa_to_upper_gva(uint64_t gpa)
 static inline int gva_to_memreg_idx(km_gva_t addr)
 {
    addr = gva_to_gpa(addr);   // adjust for gva in the top part of VA space
-   if (addr <= machine.guest_mid_physmem) {
-      return MEM_IDX(addr);
+   if (addr > machine.guest_mid_physmem) {
+      return machine.last_mem_idx - MEM_IDX(machine.guest_max_physmem - addr - 1);
    }
-   return machine.last_mem_idx - MEM_IDX(machine.guest_max_physmem - addr - 1);
+   return MEM_IDX(addr);
```
paulpopelka, line 188: This change seems to just rearrange code.
There is no logical difference.
Does this give some small performance advantage?

km/km_mem.h
```diff
@@ -116,11 +122,27 @@ static const uint64_t GUEST_ARG_MAX = 32 * KM_PAGE_SIZE;
...
+ * PIE, static or dynamic.
+ *
+ * Per https://stackoverflow.com/questions/61561331/why-does-linux-favor-0x7f-mappings,
+ * Minimum address for ET_DYN, if we take a look at ELF_ET_DYN_BASE, we see that it is architecture
+ * dependent and on x86-64 it evaluates to:
+ * ((1ULL << 47) - (1 << 12)) / 3 * 2 == 0x555555554aaa
  */
-static const km_kma_t KM_USER_MEM_BASE = (void*)0x100000000000ul;   // 16TiB
+#ifndef KM_GPA_AT_16T
+static const km_kma_t KM_USER_MEM_BASE = (void*)0x0;
```
paulpopelka, line 142: Can we find a more automatic way to propagate changes to this symbol's value to kkm?
Or, can km query kkm to find out if the version that supports the new value is available?

sv641 (dismissed):

paulpopelka (approved):

PR #1535, opened 2022-03-16
Remove dependency on libelf - process elf 64 ourselves

johnmuth81:
libelf is a 20K line library that we use a time amout of to read elf files.
We don't use it for writing elf files (core/snapshot). This commit
moves all ELF processing into KM and removes any dependency on libelf.

### DIFF
--- a/km/load_elf.c
+++ b/km/load_elf.c
@@ -85,66 +85,47 @@ static void load_extent(int fd, const GElf_Phdr* phdr, km_gva_t base)
    }
 }
 
-static void km_find_dlopen(Elf* e, km_gva_t adjust)
+static void km_find_dlopen(km_elf_t* e, km_gva_t adjust)
 {
-   for (Elf_Scn* scn = NULL; (scn = elf_nextscn(e, scn)) != NULL;) {
-      GElf_Shdr shdr;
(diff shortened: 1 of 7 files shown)


km/load_elf.c
```diff
@@ -263,3 +240,97 @@ uint64_t km_load_elf(km_elf_t* e)
...
+
+   // Find Symbol table and String Table Indexes
+   elf->symidx = elf->stridx = -1;
+   for (int i = 0; i < elf->ehdr->e_shnum; i++) {
+      if (elf->shdr[i].sh_type == SHT_SYMTAB) {
+         elf->symidx = i;
+      }
+      if (elf->shdr[i].sh_type == SHT_STRTAB && i != elf->ehdr->e_shstrndx) {
+         elf->stridx = i;
+      }
```
johnmuth81, line 308: This is actually available in the `sh_link` field in the `SHT_SYMTAB` section header.

gnode1 (dismissed):
LGTM. I didn't go too deep into the parsing but I did some cursory review there as well. There are several warnings as to multiplication overflow which I'd assume you want fixed. And of course I couldn't resist on nits :-).

gnode1 (approved):

PR #1541, opened 2022-03-28
change close order during fork and exec

sv641:
close vcpus, machine and them kvm fd in this specific order.


gnode1 (approved):
LGTM, but a bit unclear why/if it even maters

sv641:
> LGTM, but a bit unclear why/if it even maters

when close is called by userland code, resources associated with machine are freed. If a vcpu is running at this time it gets messed up. 

KKM implements a refcount to handle out of order closes(vcpu and machine), but the path was never tested.
There is a memory access after free in that code about 10 instructions window.
Separate review is on kkm repo for that fix.

PR #1545, opened 2022-04-01
preserve sockinfo across dup style syscalls

### LINKED ISSUE #1544: KM dup paths lose sockinfo
Found during @smijar testing for python flask. The specific scenario is a 'duplicating' system call (in this case `fcntl(F_DUPFD_CLOEXEC)` is called with a socket as the source file descriptor. The `sockinfo` structure is not copied into the new fd.  Manifested as a SIGSEGV in flask.

johnmuth81:
The `dup` family of system calls were not preserving `sockinfo` inside KM. The new pyflask tests uncovered this bug.

Fixes #1544

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Kontain Inc
+ * Copyright 2022 Kontain Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


km/km_filesys.c
```diff
@@ -482,12 +482,23 @@ uint64_t km_fs_fcntl(km_vcpu_t* vcpu, int fd, int cmd, uint64_t arg)
...
       if (ret >= 0) {
          km_file_t* file = &km_fs()->guest_files[fd];
-         ret = km_add_guest_fd_internal(vcpu,
-                                        ret,
-                                        km_guestfd_name(vcpu, fd),
-                                        (cmd == F_DUPFD) ? 0 : O_CLOEXEC,
-                                        file->how,
-                                        ops);
+         if (file->sockinfo != NULL) {
+           ret = km_add_socket_fd(vcpu,
```
paulpopelka, line 486: If the duped socket has a local address bound, this function does not add the bind address to the new km_fd_socket_t.

tests/socket_test.c
```diff
@@ -79,12 +79,18 @@ TEST test_sockopt()
...
+int dup_fd = 0;
+
 int TEST_PORT;
 void* tcp_server_main(void* arg)
 {
    static int rc = 0;
    int fd = *(int*)arg;
 
+   if (dup_fd != 0) {
+      fd = fcntl(fd, F_DUPFD_CLOEXEC, 0);
```
paulpopelka, line 91: You are just causing the open fd's sockinfo to be copied to the new fd.
What part of this test verifies that it is there in the new fd?

gnode1 (approved):

PR #1546, opened 2022-04-04
restore function name in log

gnode1:
`km_trace()` family was redone as function instead of macros. Unfortunately that killed function names in the traces, function name became `km_trace` instead of the function that is doing the trace. This one reverts that.


johnmuth81 (approved):
lgtm

paulpopelka (approved):
I think I approved the change that brought this macro to function conversion.

Oops.

PR #1547, opened 2022-04-04
clear state in child

sv641:
Some node tests still fail

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -2636,6 +2636,11 @@ void km_fs_fini(void)
 
 static int internal_fd = KM_START_FDS;
 
+void km_filesys_internal_fd_reset()
+{
+   internal_fd = KM_START_FDS;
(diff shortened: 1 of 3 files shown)


paulpopelka (commented):
I think there will need to be a test where the payload creates the maximum number of threads (257 I think).  The thread limit is the per process vcpu limit.
The threads just wait to be told to terminate.
Then one thread will fork().
The child process will create the maximum number of threads too.
Wait for all threads to start.
Once all threads are started, ask them to stop.  Then wait for them all to terminate using pthread_join().
If any thread create fails the test fails.

This will verify that the km reserved fd's are there.
I'm not sure how to test if vm_vcpu_run_cnt is correct after fork().

sv641:
> 

I verified vm_vcpu_run_cnt manually.

gnode1 (approved):

PR #1548, opened 2022-04-05
WIP: prototype glibc dynamic library support

johnmuth81:
These are the changes from the NVIDIA support effort to support binaries linked with glibc ld.so under KM. The changes required were:

1) support fixed mmap in low memory. This is hacked by moving `brk` unmap will just puke.
2) Don't try to load all ELF files at `GUEST_MEM_START_VA`. Do what the ELF file says. So what if a few MB is wasted?
3) Explicitly clear residual memory in mmap'ed page to support `ld.so` assumption.

I'm making this a PR to discuss with the group.

### DIFF
--- a/km/km_mmap.c
+++ b/km/km_mmap.c
@@ -575,6 +575,28 @@ km_mmap_add_region(km_gva_t gva, size_t size, int prot, int flags, int hostfd, m
    return ret;
 }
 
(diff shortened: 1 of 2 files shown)


johnmuth81:
obsolete due to #1561

PR #1553, opened 2022-04-18
Ez leka/krun separation

ez-leka:
Krun-static build job now stores persistent artifact. KRUN  is only built if submodule  have been changed, otherwise the exiting artifact from previous build is retrieved and used. 

Job uses new custom action to retrieve and maintain persistent artifact:  ez-leka/persistent-artifact@v1. This custom action will attempt to retrieve  artifact. If found, will download it just like standard download-artifact action and will set output variable of the step 'artifact-status' to 'available'. It will also re-upload artifact to prevent if from expiring. If there is no requested artifact, it will set 'artifact-status' to 'not-found' and return without error. 

'artifact-status' is then used as a condition to determine whether to start run build or if a buid is already available. 

For KRUN-static purposes the name of the persistent artifact is derived from submodule SHA. That way , artifact name will automatically change when there are changes to the KRUN and persist otherwise.

### DIFF
--- a/.github/workflows/km-ci-workflow.yaml
+++ b/.github/workflows/km-ci-workflow.yaml
@@ -183,8 +183,7 @@ jobs:
   # parallel with the rest of KM build and test.
   krun-static-build:
     name: Build static KRUN binary
-    runs-on: ${{ fromJSON(needs.start-runners.outputs.run-ons)['ec2'] }}
-    needs: start-runners
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


gnode1 (dismissed):

.github/workflows/km-ci-workflow.yaml
```diff
@@ -183,19 +183,32 @@ jobs:
...
   krun-static-build:
     name: Build static KRUN binary
-    runs-on: ${{ fromJSON(needs.start-runners.outputs.run-ons)['ec2'] }}
-    needs: start-runners
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+      - name: Set env
```
gnode1, line 191: Some comments would be good here. Like we use SHA as a name and fetch the artifact from prior runs if there is one. If not, build afresh and then upload with the SHA name

gnode1 (approved):

PR #1554, opened 2022-04-18
ELF load cleanup

johnmuth81:
No need to buffer Phdr, Shdr, and symbol table string table. stdio is capable of doing that.

### DIFF
--- a/km/load_elf.c
+++ b/km/load_elf.c
@@ -93,26 +93,42 @@ static void load_extent(int fd, const Elf64_Phdr* phdr, km_gva_t base)
 
 static void km_find_dlopen(km_elf_t* e, km_gva_t adjust)
 {
-   if (e->symidx == -1 || e->stridx == -1) {
-      return;
+   int found = 0;
+   Elf64_Shdr sym_shdr;
+   Elf64_Shdr str_shdr;
+   for (int i = 0; i < e->ehdr.e_shnum; i++) {
+      if (km_elf_get_shdr(e, i, &sym_shdr) != 0) {
+         km_err(2, "Cannot get section header %d", i);
+      }
+
+      if (sym_shdr.sh_type == SHT_SYMTAB) {
+         if (km_elf_get_shdr(e, sym_shdr.sh_link, &str_shdr) != 0) {
+            km_err(2, "Cannot get section header %d", sym_shdr.sh_link);
+         }
+         found = 1;
+         break;
+      }
+   }
+   if (found == 0) {
+      km_err(2, "Cannot find symbol table for DLOPEN");
    }
 
-   int nsym = e->shdr[e->symidx].sh_size / e->shdr[e->symidx].sh_entsize;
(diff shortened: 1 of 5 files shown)


km/load_elf.c
```diff
@@ -93,26 +93,42 @@ static void load_extent(int fd, const Elf64_Phdr* phdr, km_gva_t base)
...
+      if (sym_shdr.sh_type == SHT_SYMTAB) {
+         if (km_elf_get_shdr(e, sym_shdr.sh_link, &str_shdr) != 0) {
+            km_err(2, "Cannot get section header %d", sym_shdr.sh_link);
+         }
+         found = 1;
+         break;
+      }
+   }
+   if (found == 0) {
+      km_err(2, "Cannot find symbol table for DLOPEN");
```
gnode1, line 113: table? Or just symbol?

km/load_elf.c
```diff
@@ -327,18 +316,35 @@ km_close_elf_file(km_elf_t* elf)
...
-      }
-      if (elf->phdr != NULL) {
-         free(elf->phdr);
-         elf->phdr = NULL;
-      }
-      if (elf->shdr != NULL) {
-         free(elf->shdr);
-         elf->shdr = NULL;
-      }
-      free(elf);
```
gnode1, line 342: Don't we need this free?

gnode1 (commented):
I suspect one of the removed `free()` calls is actually needed.

Other than that, and two code scanning alerts about potential overflow, looks good.

km/load_elf.c
johnmuth81: Yes. re-added.

km/load_elf.c
```diff
@@ -270,50 +282,27 @@ km_open_elf_file(const char *path)
    }
 
    // Read ELF header
-   elf->ehdr = (Elf64_Ehdr *) malloc(sizeof(Elf64_Ehdr));
-   int nread = fread(elf->ehdr, 1, sizeof(Elf64_Ehdr), elf->file);
+   int nread = fread(&elf->ehdr, 1, sizeof(Elf64_Ehdr), elf->file);
```
sv641, line 285: 2and 3rd params should be swapped.
Current code will work, but is confusing.

sv641 (approved):

km/load_elf.c
johnmuth81: Actually this is a conscious choice since I want the return value to be the number of bytes read, not the number of records (see the man page). fread/fwrite are kind of weird since the return value is the number of 'items' transferred. I've always found it easier to make the item size '1' and deal in bytes just like read/write. I guess it's just laziness on my part.

km/load_elf.c
johnmuth81: We are searching the section headers for the symbol table and it's associated string table. We look for the symbol itself further down.

gnode1 (approved):

PR #1555, opened 2022-04-18
Cleanup withpacker and k8s targets and docs. Speedup km build

gnode1:
The km build speedup is simply by running things in parallel (`make -j`) where it makes sense.

Cleanup is about remove `-withpacker` and `-withk8s` targets, and update the docs to reflect on-demand runners.


sv641 (approved):

paulpopelka:
Ok.  I looked at it.

PR #1558, opened 2022-04-25
recover brk/tbrk from snapshot

### LINKED ISSUE #1539: Pass brk and tbrk in snapshot instead of determining them by inspection of phdr
km_snapshot.c lines 161, 216, 397 we tell upper VA addresses from lower ones, assuming upper addressees located high in the address space. PR #1534 implements gva and kma equivalency, which makes this check impossible. Right now in that PR we use hack to tell upper from lower.

Proposal - pass values of brk and tbrk in the snapshot file, and use them. This allows to avoid inspecting the phdr and makes it easy to tell upper fro lower.

johnmuth81:
fixes #1539

### DIFF
--- a/km/km.h
+++ b/km/km.h
@@ -362,9 +362,6 @@ typedef struct km_machine {
 
    uint64_t xcr0;
    int overcommit_memory;   // controls how we request memory for payload from Linux
-
-   km_gva_t snap_rcvr_brk;
-   km_gva_t snap_rcvr_tbrk;
 } km_machine_t;
 
 extern km_machine_t machine;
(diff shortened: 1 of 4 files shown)


km/km.h
```diff
@@ -362,6 +362,9 @@ typedef struct km_machine {
 
    uint64_t xcr0;
    int overcommit_memory;   // controls how we request memory for payload from Linux
+
+   km_gva_t snap_rcvr_brk;
```
gnode1, line 366: Do we really need to keep these in here - looks like we only need them for a very short time recovering from the snap?

gnode1 (approved):

PR #1561, opened 2022-04-27
dynamic glibc plus some tests

johnmuth81:
- Support programs linked against GLIBC dynamic linker.
- Add tests

Currently the GBD server does not work with programs linked against the GLIBC dynamic linker. This is due to the linker
and LIBC being separate files in GLIBC vs one file in MUSL.

### DIFF
--- a/km/load_elf.c
+++ b/km/load_elf.c
@@ -38,8 +38,7 @@ km_payload_t km_dynlinker;
 static void map_program_section(int fd, void* buf, size_t count, off_t offset)
 {
    if (count > 0) {
-      if (mmap(buf, count, PROT_WRITE, MAP_PRIVATE | MAP_FIXED, fd, offset) ==
-          MAP_FAILED) {
+      if (mmap(buf, count, PROT_WRITE, MAP_PRIVATE | MAP_FIXED, fd, offset) == MAP_FAILED) {
          km_err(2, "error mmap elf");
       }
       if (count != roundup(count, KM_PAGE_SIZE)) {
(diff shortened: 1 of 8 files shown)


tests/.gitignore
```diff
@@ -2,6 +2,7 @@ km-bats-*.*
 gdb.txt
 .native*
 *.fedora
+*.fedora.glibc
```
gnode1, line 5: .fedora is also glibc, but static. Maybe call it .fedora.dyn or something?

tests/Makefile
```diff
@@ -81,7 +81,7 @@ LDFLAGS := ${DEBUG}
 
 DEPS = ${SRC_C:%.c=%.d} ${SRC_SHR:%.c=%.d} ${SRC_CPP:%.cpp=%.d} ${HELPER_SRC:%.c=%.d}
 
-SUFFIXES := .km .kmd .km.so .alpine.km .alpine.kmd .${DTYPE}
+SUFFIXES := .km .kmd .km.so .alpine.km .alpine.kmd .${DTYPE} .${DTYPE}.glibc
```
gnode1, line 84: Same as above on .glibc. Other laces below

tests/km_core_tests.bats
```diff
@@ -42,6 +42,7 @@ todo_alpine_static='dl_iterate_phdr gdb_forkexec'
 
 # glibc native
 not_needed_glibc_static='cpuid setup_link setup_load gdb_sharedlib readlink_argv km_identity dlopen exec_sh'
+not_needed_glibc_dynamic='cpuid setup_link setup_load gdb_sharedlib readlink_argv km_identity dlopen exec_sh mem_brk'
```
gnode1, line 45: Just curious, why mem_brk is excluded?

km/load_elf.c
```diff
@@ -230,25 +239,40 @@ uint64_t km_load_elf(km_elf_t* e)
...
     *
     * We load DYNAMIC starting at GUEST_MEM_START_VA, adjust will tell how much shift there was. For
     * loading PIE that doesn't matter, but we need to keep it for coredump and such. For dynamic
     * linking adjust is passed to the dynamic linker so it know how to do relocations.
     */
-   if (km_guest.km_dynamic_vaddr != 0 && km_guest.km_dynamic_len != 0) {
+   if (e->ehdr.e_type == ET_EXEC) {
+      km_infox(KM_TRACE_LOAD, "type is ET_EXEC");
+   } else if (e->ehdr.e_type == ET_DYN) {
+      km_infox(KM_TRACE_LOAD, "type is ET_DYN");
```
gnode1, line 270: Looking at my system (Fedora 35) I see `/bin/cat` being PIE and `/bin/gcc` being regular dynamic exec:

/bin/gcc: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=1afd7874e01ce0a1ecca42cc3a42a022127e8d50, for GNU/Linux 3.2.0, stripped
/bin/cat: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=b5f22ffbc60b2993dc4f138c5a052e994d404cb7, for GNU/Linux 3.2.0, stripped

However I see them both showing only DYNAMIC header. In fact I was looking for something with ET_EXEC and found none. What am I missing?

With that in mind, I don't have a clear picture how this covers all cases: static, regular dynamic, dynamic PIE, and static PIE.

johnmuth81: Here is a comment I found in linux `fs/binfmt_elf.c` which I found helpful:
```
 * There are effectively two types of ET_DYN
 * binaries: programs (i.e. PIE: ET_DYN with INTERP)
 * and loaders (ET_DYN without INTERP, since they
 * _are_ the ELF interpreter). The loaders must
 * be loaded away from programs since the program
 * may otherwise collide with the loader (especially
 * for ET_EXEC which does not have a randomized
 * position).
```

tests/km_core_tests.bats
johnmuth81: No good reason. Moved to a 'todo'

gnode1 (approved):

PR #1562, opened 2022-04-28
Automatic ENOTSUP for unimplemented system calls

johnmuth81:
This is a proposal to change the default for handling unimplemented systems calls from the current behavior, killing the guest with a `SIGSYS` signal, to returning `-ENOSYS`. (Not Supported). The reasoning for this is experiments in bringing up new apps that attempt to use system calls but are tolerant of ENOSYS. They would have come right up with this implemented where with current behavior, each new system call required a hander to return `-ENOSYS`.

I've included a command line option to go back to the old behavior.

### DIFF
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -481,6 +481,11 @@ static void km_vcpu_exit_all(km_vcpu_t* vcpu)
    // TODO - consider an unforced solution
    if (km_vcpu_run_cnt() > 1) {
       km_infox(KM_TRACE_VCPU, "Forcing exit_group() without cleanup");
+      /*
+       * Even though we are forcing exit, do a hcalls fini since that will report
+       * statistics is requested.
+       */
+      km_hcalls_fini();
       exit(machine.exit_status);
    }
    km_vcpu_stopped(vcpu);
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -37,6 +37,7 @@
 
 int vcpu_dump = 0;
 int km_collect_hc_stats = 0;
+int kill_unimpl_hcall = 0;
 
 #define fx "VCPU %d RIP 0x%0llx RSP 0x%0llx CR2 0x%llx "
 
@@ -412,14 +413,14 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
          km_post_signal(vcpu, &info);
          return -1;
       }
-      if (km_hcalls_table[hc] == NULL) {
+      if (kill_unimpl_hcall != 0 && km_hcalls_table[hc] == NULL) {
          km_warnx("Unimplemented hypercall %d (%s)", hc, km_hc_name_get(hc));
          siginfo_t info = {.si_signo = SIGSYS, .si_code = SI_KERNEL};
          km_post_signal(vcpu, &info);
          return -1;
       }
-      vcpu->hypercall = hc;
    }
+   vcpu->hypercall = hc;
    // We assume hypercall args are built are on stack in the guest, but nothing in km depends on this.
    ga = (km_gva_t)km_hcargs[HC_ARGS_INDEX(vcpu->vcpu_id)];
    km_hc_args_t* ga_kma = km_gva_to_kma(ga);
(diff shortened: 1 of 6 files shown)


km/km_vcpu_run.c
```diff
@@ -435,8 +436,15 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
...
    }
-   km_hc_ret_t ret = km_hcalls_table[hc](vcpu, hc, ga_kma);
-   *hc_ret = ga_kma->hc_ret;
+   km_hc_ret_t ret = HC_CONTINUE;
+   if (km_hcalls_table[hc] != NULL) {
+      ret = km_hcalls_table[hc](vcpu, hc, ga_kma);
+      *hc_ret = ga_kma->hc_ret;
+   } else {
+      km_infox(KM_TRACE_HC, "Unimplemented hypercall %d (%s)", hc, km_hc_name_get(hc));
+      ga_kma->hc_ret = (uint64_t) -ENOTSUP;
```
paulpopelka, line 445: Maybe we could return ENOSYS so they know it is the system call that is not supported, not some unsupported argument to the system call.

km/km_vcpu_run.c
```diff
@@ -37,6 +37,7 @@
 
 int vcpu_dump = 0;
 int km_collect_hc_stats = 0;
+int kill_unimpl_hcall = 0;
```
sv641, line 40: move to machine.
Slowly move to no gobal variables.

km/km_vcpu_run.c
```diff
@@ -435,8 +436,15 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
    if (km_collect_hc_stats != 0) {
       clock_gettime(CLOCK_MONOTONIC, &start);
    }
-   km_hc_ret_t ret = km_hcalls_table[hc](vcpu, hc, ga_kma);
-   *hc_ret = ga_kma->hc_ret;
+   km_hc_ret_t ret = HC_CONTINUE;
+   if (km_hcalls_table[hc] != NULL) {
+      ret = km_hcalls_table[hc](vcpu, hc, ga_kma);
+      *hc_ret = ga_kma->hc_ret;
```
paulpopelka, line 442: Move the assignment to *hc_ret out of the if then else since both sides do it and both sides have ga-kma->hc_ret already set.

km/km_vcpu_run.c
gnode1: Make it an environment variable as well, easier to set in various environments.

gnode1:
This is a trade of between visibility and convenience. On one hand we'd like failures to be explicit and obvious, so we can investigate right away. On the other hand well behaved programs might be able to continue working correctly even in absence of some syscalls. Or not - forgetting to check the result and going crazy.

km/km_vcpu_run.c
```diff
@@ -413,7 +414,7 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
          km_post_signal(vcpu, &info);
          return -1;
       }
-      if (km_hcalls_table[hc] == NULL) {
+      if (kill_unimpl_hcall != 0 && km_hcalls_table[hc] == NULL) {
```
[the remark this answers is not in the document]
johnmuth81: I did it this way because I assumed since `kill_unimpl_hcall == 0` is the default, the first expression would be false most of the time and we'd save the indirection for `km_hcalls_table[hc] == NULL`. When `kill_unimpl_hcall != 0` we are debugging and don't care about performance.

What's your reasoning for the opposite order?

km/km_vcpu_run.c
johnmuth81: OK. I can do ENOSYS instead of ENOTSUP.

km/km_vcpu_run.c
johnmuth81: OK

km/km_vcpu_run.c
johnmuth81: I'm going to create an environment variable this one, but I'm going to hold off putting it into `machine`.

Our current config stuff is a bit of a mess. I'd like to see a more generalized `config` service that could deal with command line options and environment variables. I see it presenting a common way to inquire if an option is enabled and/or an optional argument is present.

I'd also like to review what we have now and see if it all still makes sense.

km/km_vcpu_run.c
johnmuth81: I'm changing it back since it seems some important user land frameworks handle `ENOTSUP` better than `ENOSYS`.

gnode1 (approved):

PR #1564, opened 2022-04-29
TRIVIAL: clang-format, use `mkdir -p xxx` in lieu of `[ -d xxx ] || mkdir xxx`

gnode1:


sv641 (approved):

lib/libkontain/Makefile
```diff
@@ -28,8 +28,8 @@ INCLUDES := ${TOP}/include
 include ${TOP}/make/actions.mk
 
 all: libkontain.h
-	[ -d ${KM_OPT_INC} ] || mkdir ${KM_OPT_INC}
+	mkdir -p ${KM_OPT_INC}
 	cp libkontain.h ${KM_OPT_INC}
-	[ -d ${KM_OPT_LIB} ] || mkdir ${KM_OPT_INC}
+	mkdir -p ${KM_OPT_INC}
```
paulpopelka, line 33: I think this should be creating ${KM_OPT_LIB}, not *_INC

Looks like this bug has been there a while.

PR #1574, opened 2022-05-04
remove symlink and shell tricks from km

gnode1:
As discussed, undo the symlink tricks with .km. Make tests and payload tests work with this.

Big question to the group - did I cut too much?


paulpopelka:
Looks like the exec test for glibc_static and glibc_dynamic are failing.

gnode1:
> 

These tests were never turned on for glibc, and it turns out they were broken. In addition, the problem actually doesn't exist with newer version of glibc. So I disabled them for glibc tests.

johnmuth81 (approved):
My favorite kind of change, lots code removed.

payloads/node/scripts/test-run.sh
```diff
@@ -50,13 +50,14 @@ if [[ "$1" == "test" || "$1" == "test-all" ]]; then
...
-#    NODETOP=$5
-#    BUILD=$6
-# 	cd ${NODETOP}
-#    python tools/test.py -J --mode=`echo -n ${BUILD} | tr '[A-Z]' '[a-z]'` --skip-tests=`cat ../skip_* ../${PLATFORM_ID}_skip | tr -s '\n ' ','` default addons js-native-api node-api
-#    echo "Tests are Successful"
-# fi
+if [[ "$1" == "test-all" ]]; then
+   NODETOP=$5
+   BUILD=$6
+   echo -e "#!/bin/bash\n$(realpath ${KM_BIN}) $(realpath ${NODETOP}/out/${BUILD}/node.km) \$*\n" > /tmp/nodexxx && chmod +x /tmp/nodexxx
```
paulpopelka, line 56: Consider using mktemp to create a temp file instead of hardcoded /tmp/nodexxx.

Can you remove the temp file after the test is done?

km/km_exec.c
```diff
@@ -529,74 +359,19 @@ char** km_exec_build_argv(char* filename, char** argv, char** envp)
 {
```
paulpopelka, line 359: See above: I think we can get rid of km_is_shell_path() and km_is_env_path() and km_one_of() and 
static const_string_t SHELL_PATHS[] =
static const_string_t ENV_PATHS[] =

And the function prototypes in km/km_exec.h

And remove the comment about trying to run the shell in the block comment above.

Hope I'm not missing some reason to keep these around.

km/km_main.c
```diff
@@ -636,30 +613,9 @@ int main(int argc, char* argv[])
...
-       */
-      if ((pl_name = km_traverse_payload_symlinks((const char*)getauxval(AT_EXECFN))) == NULL) {
-         pl_name = km_traverse_payload_symlinks((const char*)argv[0]);
-      }
-   } else {
-      pl_name = NULL;
-   }
+   char* pl_name = NULL;
+
    km_trace_setup(argc, argv, pl_name);   // setup trace settings as early as possible
```
paulpopelka, line 618: Since pl_name is always null now, there are probably changes needed in km_trace_setup.
At the least we can get rid of the pl_name variable.

paulpopelka: Not sure about getting rid of pl_name entirely, looks like we pass it to km_parse_args().

paulpopelka: Looks like km_parse_args() could use some code removal too if the pl_name argument is always null.
But... I think this change is pretty big and don't want it to get bigger.

paulpopelka (commented):
Other than removing more code that has been made unused by this change, the change looks ok.

Do you want to remove more code or just get this in and leave additional cleaning for another change?

paulpopelka (approved):

paulpopelka (approved):

PR #1575, opened 2022-05-06
km changes related to KKM implementing 128 byte redzone according to …

sv641:
…system V ABI

### DIFF
--- a/km/km_kkm.c
+++ b/km/km_kkm.c
@@ -81,3 +81,19 @@ int km_kkm_get_xstate(km_vcpu_t* vcpu, kkm_xstate_t* kx)
    }
    return retval;
 }
+
+/*
+ * Adjust stack depending on syscall or out instruction.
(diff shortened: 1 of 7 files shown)


sv641:
needs kkm review and rebase before tests will pass.

https://github.com/kontainapp/kkm/pull/16

km/km_kkm.c
```diff
@@ -81,3 +81,17 @@ int km_kkm_get_xstate(km_vcpu_t* vcpu, kkm_xstate_t* kx)
    }
    return retval;
 }
+
+/*
+ * depending on syscall or out stack adjustment changes
```
gnode1, line 86: I think it warrants an explanation why the adjustment is different for these two cases

sv641: added comments.

gnode1 (approved):

PR #1582, opened 2022-05-11
handle null file name in utimensat syscall

sv641:
one test closer


sv641:
> Test would be nice

added.

PR #1583, opened 2022-05-11
use same virt-device after exec

sv641:


gnode1 (commented):
I'm guessing what this change addresses - running on a machine with both kvm and kkm enabled, exec needs to be told which one to use. An explanation would be nice.

Also test?

km/km_exec.c
```diff
@@ -368,15 +368,32 @@ char** km_exec_build_argv(char* filename, char** argv, char** envp)
...
       return NULL;
    }
-   nargv[0] = km_get_self_name();   // km exec
-   nargv[1] = pl_name;
-   memcpy(nargv + 2, argv + 1, sizeof(char*) * (argc - 1));
+   int new_argc_index = 0;
+   nargv[new_argc_index++] = km_get_self_name();   // km exec
+   if (km_machine_init_params.vdev_name != NULL) {
+      int string_len = strlen("--virt-device=") + strlen(km_machine_init_params.vdev_name) + 1;
+      char* string_ptr = calloc(string_len, sizeof(char));
```
paulpopelka, line 385: calloc() seems pointless since you overwrite the string completely with snprintf() below.

km/km_exec.c
```diff
@@ -368,15 +368,32 @@ char** km_exec_build_argv(char* filename, char** argv, char** envp)
...
       free(argv);
       return NULL;
    }
-   nargv[0] = km_get_self_name();   // km exec
-   nargv[1] = pl_name;
-   memcpy(nargv + 2, argv + 1, sizeof(char*) * (argc - 1));
+   int new_argc_index = 0;
+   nargv[new_argc_index++] = km_get_self_name();   // km exec
+   if (km_machine_init_params.vdev_name != NULL) {
+      int string_len = strlen("--virt-device=") + strlen(km_machine_init_params.vdev_name) + 1;
```
paulpopelka, line 384: Maybe make a preprocessor symbol for "--virt-device=" since we use it twice here.
It helps people know they are the same.

km/km_exec.c
sv641: changed to malloc

paulpopelka (approved):

PR #1584, opened 2022-05-12
Clean FP state check in snapshot test

gnode1:
Trying this on Fedora 36 - gcc version 12.1.1. Turns out the structure assignment uses xmm0 register. Since the test value was assigned to the xmm0 register before the structure assignment, by the time snapshot is done the value in xmm0 is erased, and the further check after restore fails.

Moved the structure assignment to before setting xmm0 register test value, which is now done right before the snapshot call. No changes in km or anywhere else, just in the test program itself.

### DIFF
--- a/tests/snapshot_test.c
+++ b/tests/snapshot_test.c
@@ -232,20 +232,20 @@ void* thread_main(void* arg)
    double a1 = 99.99;
    register double a2 asm("xmm7") = 101.202;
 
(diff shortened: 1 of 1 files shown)


sv641 (approved):

sv641:
Looks like now compiler owns xmm registers.
It used to be application owned.

PR #1587, opened 2022-05-14
Ez leka/k8s release

ez-leka:
Added 2 targets to top level makefile to automate updating and properly tagging deploy yamls:
1. deploy-config - builds yaml files using RELEASE_TAG specified. If none, v0.1.-test default is used
2. release-prep - depends on deploy-config. After all deployy yaml files are built, tags release using RELEASE_TAG specified. If none, v0.1.-test default is used

Also, I have  changed release document to reflect new way of creating release. Manual manipulation of k8s yaml files is no longer required. Please review and accept document changes as well.


smijar (approved):
Looks great. Thanks a lot.

PR #1588, opened 2022-05-14
Support gcc-12 compiled dlopen

gnode1:
Switching to Fedora 36 brought gcc 12, which compiles `dlopen()` into slightly different assembly - essentially the instruction in question is shifted by 3 bytes in memory. So inspect both locations and go accordingly.

### DIFF
--- a/km/km_musl_related.c
+++ b/km/km_musl_related.c
@@ -71,8 +71,10 @@ int km_link_map_walk(link_map_visit_function_t* callme, void* visitargp)
    km_gva_t linkmapheadp_gva;
    int rc = 0;
 
-   static const uint64_t KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR = 0x11;
-   static const uint64_t KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR_gcc_12 = 0x14;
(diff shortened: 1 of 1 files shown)


km/km_musl_related.c
```diff
@@ -68,8 +71,13 @@ int km_link_map_walk(link_map_visit_function_t* callme, void* visitargp)
    km_gva_t linkmapheadp_gva;
    int rc = 0;
 
-#define KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR 0x11
-#define KM_DLOPEN_OFFSET_TO_POST_LOAD_HEAD_INSTR 0x18
+   static const uint64_t KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR = 17;
+   static const uint64_t KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR_gcc_12 = 20;
+   static const uint64_t KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR_O0 = 45;
```
paulpopelka, line 76: Can you briefly describe the situation where this would happen.
Is optimization disabled?

paulpopelka (approved):
Other than explaining what situation KM_DLOPEN_OFFSET_TO_LOAD_HEAD_INSTR_O0 is for it looks fine.
I assume O 0 is gcc optimization level zero.  Would help to add a comment.

PR #1589, opened 2022-05-16
save km core in CI

gnode1:
There are two changes here. First we change the `kernel.core_pattern` in the L0 base image to `core_%P`, so the km core files are written in the test directory in case km crashed. And second, we keep container after the run completes `DOCKER_RUN_CLEANUP=` so we can copy the files from it for inspection.

### DIFF
--- a/.github/workflows/km-ci-workflow.yaml
+++ b/.github/workflows/km-ci-workflow.yaml
@@ -352,7 +352,7 @@ jobs:
           make -C payloads -j pull-testenv-image
 
       - name: KM with KVM Test - Azure
-        run: make -C tests test-withdocker DOCKER_INTERACTIVE=
+        run: make -C tests test-withdocker DOCKER_INTERACTIVE= DOCKER_RUN_CLEANUP=
         timeout-minutes: 20
 
       - name: KM with KVM Payloads Test - Azure
@@ -462,7 +462,7 @@ jobs:
           make -C payloads -j pull-testenv-image
 
       - name: KM with KKM Test - AWS
-        run: make -C tests test-withdocker HYPERVISOR_DEVICE=/dev/kkm DOCKER_INTERACTIVE=
+        run: make -C tests test-withdocker HYPERVISOR_DEVICE=/dev/kkm DOCKER_INTERACTIVE= DOCKER_RUN_CLEANUP=
         timeout-minutes: 20
 
       - name: KM with KKM Payloads Test - AWS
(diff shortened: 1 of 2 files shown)


paulpopelka (commented):
This comment is mostly to clarify my understanding of how this DOCKER_RUN_CLEANUP works.
DOCKER_RUN_CLEANUP is put on some "docker run" and "podman run" command lines.  By default its value is "--rm".  So the container is destroyed after the container is completed.
This change switches "make ... test-withdocker ..." in the ci workflow to keep the container around after the test completes.
In addition core files from processes in these containers now have a name with the pid of the faulted process.  The pid is the one as seen from outside the container, not the pid inside the container.
To look at the core we need to login to the runner (I have not supplied my public key yet).
Is the runner suspended?
I don't know what directory the core file will be in.
The runners are cleaned up everyday at 2am.  So we need to look at the cores promptly.

How far into the weeds have I driven?

The change looks fine if my opinion matters.

gnode1:
> This comment is mostly to clarify my understanding of how this DOCKER_RUN_CLEANUP works. DOCKER_RUN_CLEANUP is put on some "docker run" and "podman run" command lines. By default its value is "--rm". So the container is destroyed after the container is completed. This change switches "make ... test-withdocker ..." in the ci workflow to keep the container around after the test completes. In addition core files from processes in these containers now have a name with the pid of the faulted process. The pid is the one as seen from outside the container, not the pid inside the container. To look at the core we need to login to the runner (I have not supplied my public key yet). Is the runner suspended? I don't know what directory the core file will be in. The runners are cleaned up everyday at 2am. So we need to look at the cores promptly.
> 

The runners for successful runs are destroyed right away. For the failed tests the runners are kept around for 7 days. We check for the old runners every night, removing those that are more than 7 days old.

Steps to look at the failed tests:
- Identify the VM. First Azure vs EC2. The VM runner names look like `runner-KM-CI-Pipeline-1526-2`. `KM-CI-Pipeline` is workflow name, `1526` is pipeline run number. `-2` at the end is attempt number if more than 1. On EC2 the names are like: `ec2-runner-KM-CI-Pipeline-1526-2`
- Start the VM. Wait for the public IP number to show, then connect `ssh kontain@IP_number` (yes you do need public key in the repo for that).
- Inside the VM `docker ps -a` will show stopped container with the tests. A command `docker cp container_ID:/home/appuser/km/tests .` will copy everything from tests directory. Or you could `docker restart container_ID` followed by `docker exec -it container_ID bash` to get inside.



> How far into the weeds have I driven?
> 
> The change looks fine if my opinion matters.

paulpopelka (approved):

PR #1594, opened 2022-05-27
minor memory leaks

gnode1:
Change to make to build km.dyn - valgrind doesn't work very well with static executables. Fix a couple of minor memory leaks


paulpopelka (commented):
Has valgrind complained about km_payload_name?
I don't think we free the string km_payload_name points to.

gnode1:
> Has valgrind complained about km_payload_name? I don't think we free the string km_payload_name points to.

Actually we do free it. It's a roundabout way but we do. First `km_open_elf_file()` assigns that pointer to `elf->path` field. Then later in `km_load_elf()` it gets assigned to `km_guest.km_filename`. And finally in `km_machine_fini()` it gets freed.

paulpopelka (approved):
The memory leak fixes are fine.
Are the removed cloud/k8s/..... files related to the memory leaks?
Are the valgrind related changes in km/Makefile complete?
I tried to find a place where VALGRIND is set and didn't see it.

gnode1:
> The memory leak fixes are fine. Are the removed cloud/k8s/..... files related to the memory leaks? 

No, they are not. 

> Are the valgrind related changes in km/Makefile complete? I tried to find a place where VALGRIND is set and didn't see it.

Kind of but not really. I simply call `make -j VALGRIND=1` for now. When I have the valgrind changes complete that call will be implemented some place, likely in action workflow yaml

johnmuth81 (approved):

Issue #434, opened 2020-02-07
Move km cleanup code into atexit registered function


gnode1:
The km code is doing cleanup at the end of the run so that tools like valgrind report clean run with no leaks. However this cleanup is not done in cases of errors when we exit via `err()` or in case of forced exit when there are threads that are "stuck".

Linux has facility call `atexit()` that allows to register a function to be called at the exit time, so that cleanup is done unconditionally.

PR #1596, opened 2022-06-01
Eliminate forced exit

gnode1:
Fixes https://github.com/kontainapp/km/issues/434

### DIFF
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -456,37 +456,14 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
    return ret;
 }
 
-static int km_vcpu_print(km_vcpu_t* vcpu, void* skip_me)
-{
-   if (vcpu != skip_me) {
-      km_infox(KM_TRACE_VCPU, "VCPU %d still running: thread %#lx", vcpu->vcpu_id, vcpu->vcpu_thread);
-   }
-   return 0;
-}
-
 static void km_vcpu_exit_all(km_vcpu_t* vcpu)
 {
    machine.exit_group = 1;   // make sure we exit and not waiting for gdb
-   km_vcpu_pause_all(vcpu, GUEST_ONLY);
-   /*
-    * At this point there are no threads in the guest (assert at the end of km_vcpu_pause_all).
-    * However there are possibly threads in uninterruptible system calls on behalf of the guest,
-    * like futex wait. There isn't much we can do about them, so just force the exit.
-    */
-   for (int count = 0; count < 10 && km_vcpu_run_cnt() > 1; count++) {
-      km_vcpu_apply_all(km_vcpu_print, vcpu);
-      nanosleep(&_1ms, NULL);
+   km_vcpu_pause_all(vcpu, ALL);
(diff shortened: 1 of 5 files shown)


km/km_vcpu_run.c
```diff
@@ -456,37 +456,14 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
...
-      km_infox(KM_TRACE_VCPU, "Forcing exit_group() without cleanup");
-      /*
-       * Even though we are forcing exit, do a hcalls fini since that will report
-       * statistics is requested.
-       */
-      km_hcalls_fini();
-      exit(machine.exit_status);
+   km_vcpu_resume_all();
+   if (km_gdb_client_is_attached() != 0) {
+      km_gdb_notify(vcpu, GDB_KMSIGNAL_THREADEXIT);
```
paulpopelka, line 464: This doesn't seem like the right thing to do here.
We are asking the vcpu's to exit by setting machine.exit_group.
Then we tell gdb a single thread has exited.
We haven't asked the remaining payload threads to pause by sending them the KM_SIGVCPUSTOP signal.  So, I guess the payload threads keep running until they need something from the kernel and then we/km finds exit_group is set which causes the payload threads to really stop.  So km is just waiting until the threads stumble onto exit_group being set.  km does not actively try to get them to stop with this change.  Should you be doing a pthread_cancel for the remaining threads?
I need to think more about this.  The interactions between the vcpu runner and gdb and exit_group are delicate.

gnode1: Just to explain my thinking here. The intent I had was more centered on gdb thread. The gdb thread is waiting for client input or _some_ event from km. Unless we send something it will sit in the wait. This makes it exit the wait.

I didn't want to stop all vcpus because then we need to somehow make them exit that wait.

gnode1: I pushed another commit which I think addresses your concerned. Still one test - gdb_forkexec - fails pretty reliably, which one of the processes - child? - stuck in `km_gdb_main_loop()`

km/km_vcpu_run.c
```diff
@@ -598,7 +582,7 @@ static inline void km_vcpu_handle_pause(km_vcpu_t* vcpu, int hc_ret)
       km_cond_wait(&machine.pause_cv, &machine.pause_mtx);
       vcpu->state = HYPERCALL;
    }
-   km_mutex_unlock(&machine.pause_mtx);
+   pthread_cleanup_pop(1);   // km_mutex_unlock(&machine.pause_mtx);
```
sv641, line 585: what is the reason for using cleanup_pop instead of using regular mutex_unlock?

sv641 (approved):

km/km_vcpu_run.c
gnode1: We do need to do the cleanup, and the unlock. I could do `km_mutex_unlock(&machine.pause_mtx); pthread_cleanup_pop(0);` - boils down to the same thing AFAIK

sv641: Go it. Thanks.

johnmuth81 (approved):
lgtm

PR #1604, opened 2022-06-09
plug some mem leaks

gnode1:


km/km_gdb_stub.c
```diff
@@ -1408,6 +1413,10 @@ static void handle_qxfer_execfile_read(const char* packet, char* obuf)
       obuf[1 + bytes_to_deliver] = 0;
    }
    send_packet(obuf);
+   if (obuf[0] == 'l') {
+      free(gdb_execfile_name);
+      gdb_execfile_name = NULL;
```
paulpopelka, line 1418: I think you need to set gdb_execfile_len to zero here too.

paulpopelka (approved):
Looks fine except for one comment that should be fixed.

km/km_gdb_stub.c
gnode1: I guess. OTOH it seems to me the decision is based on `offset == 0`, in which case `gdb_execfile_len` is unconditionally recomputed. Am I missing something?

gnode1: The way I see it, if `gdb_execfile_name` is freed, three lines down `gdb_execfile_name` is assigned unconditionally. Of curse this depends on the client useing `offset` in a specific pattern.

I'll make the change as it certainly doesn't make anything worse. I'm just explaining the way I was thinking about it.

PR #1610, opened 2022-06-19
Couple of fixes found during valgrind work

gnode1:
Stop vcpus before memory fini. Otherwise occasionally vcpu thread blows up stepping on unmapped memory

Fix race in gdb_attach. Since valgrind makes everything run slow, gdb would attach to the process *before* all the threads are started, hence the incompatible results. Make sure guest is at the right state when attaching.

### DIFF
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -103,6 +103,13 @@ void km_vcpu_fini(km_vcpu_t* vcpu, int join_thr)
  */
 void km_machine_fini(void)
 {
+   for (int i = 0; i < KVM_MAX_VCPUS; i++) {
+      km_vcpu_t* vcpu;
+
+      if ((vcpu = machine.vm_vcpus[i]) != NULL) {
+         km_vcpu_fini(vcpu, 1);
+      }
+   }
    free(machine.auxv);
    if (km_guest.km_filename != NULL) {
       free((void*)km_guest.km_filename);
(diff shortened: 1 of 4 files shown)


km/km_cpu_init.c
```diff
@@ -103,6 +103,13 @@ void km_vcpu_fini(km_vcpu_t* vcpu, int join_thr)
  */
 void km_machine_fini(void)
 {
+   for (int i = 0; i < KVM_MAX_VCPUS; i++) {
+      km_vcpu_t* vcpu;
+
+      if ((vcpu = machine.vm_vcpus[i]) != NULL) {
+         km_vcpu_fini(vcpu, 1);
+      }
```
sv641, line 111: no need for all the vcpus.
you can do a **else { break; }**

tests/km_core_tests.bats
```diff
@@ -1345,7 +1345,11 @@ fi
 # In addition this may expose other problems such as misdeclared km arrays like km_hcargs[].
 #
 @test "threads_create($test_type): create a large number of threads that run briefly (gdb_lots_of_threads$ext)" {
+   if [ -z ${VALGRIND} ]; then
    run km_with_timeout --timeout 5s gdb_lots_of_threads_test$ext -a 2 -t 287
+   else
+   run km_with_timeout --timeout 25s gdb_lots_of_threads_test$ext -a 10 -t 287 -w
```
sv641, line 1351: indent

sv641 (approved):
waiting for some time is a little bit scary change.

gnode1:
> waiting for some time is a little bit scary change.

Not sure what you mean. This change waits for exact state of the guest

tests/gdb_lots_of_threads_test.c
```diff
@@ -160,6 +166,10 @@ int main(int argc, char* argv[])
       }
    }
 
+   while (wait_for_gdb != 0) {
+      usleep(100);
+   }
```
paulpopelka, line 171: I assume the intention here is for the user of gdb to set wait_for_gdb to zero once gdb is attached.
Maybe add a message telling the user to set wait_for_gdb = 0 to let the test proceed.

gnode1: Yes. Actually there is "-w" option to set `wait_for_gdb` to 0, and we use in in bats.

sv641:
> > waiting for some time is a little bit scary change.
> 
> Not sure what you mean. This change waits for exact state of the guest

my bad. I misunderstood.

PR #1614, opened 2022-06-21
remove exit race with gdb

gnode1:
This is pretty intricate scenario. Depending on which thread - vcpu or gdb - is faster, we occasionally end up with gdb thread waiting for the input from client while all of the vcpus are stopped. Delay sending exit even to gdb after the last vcpu truly exits.

Tested for half a day in a loop on Azure VM.

### DIFF
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -460,9 +460,6 @@ static void km_vcpu_exit_all(km_vcpu_t* vcpu)
 {
    machine.exit_group = 1;   // make sure we exit and not waiting for gdb
    km_vcpu_pause_all(vcpu, ALL);
-   if (km_gdb_client_is_attached() != 0) {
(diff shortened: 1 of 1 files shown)


.github/workflows/km-ci-workflow.yaml
```diff
@@ -97,7 +97,7 @@ jobs:
           az-tenant-id: ${{ secrets.SP_TENANT }}
           az-image: "auto-github-runner:L0BaseImage"
           az-location: "westus"
-          az-vm-size: "Standard_D4s_v3"
+          az-vm-size: "Standard_D4s_v5"
```
paulpopelka, line 100: Another extraneous change from other concurrent work?

paulpopelka (approved):
Why are these unrelated changes (az-vm-size) creeping into the PR's?
Or is it related?  Can you explain the reason in the PR description?

Does it matter?

Issue #1455, opened 2022-01-10
KM : deadlock in signal handler when back to back signals arrive


sv641:
deadlock when new signal arrives when processing signal.

problem happens when running sigsuspend test.

```
(gdb) info threads
  Id   Target Id                                   Frame 
* 1    Thread 0x174e3c0 (LWP 576483) "km"          0x000000000046a6c0 in __lll_lock_wait ()
  2    Thread 0x7f55480e4640 (LWP 576487) "vcpu-0" 0x00000000004f3cfa in __futex_abstimed_wait_common ()
(gdb) bt
#0  0x000000000046a6c0 in __lll_lock_wait ()
#1  0x000000000046e962 in pthread_mutex_lock ()
#2  0x00000000004335c7 in km_signal_lock () at /data/fedora35/km/km/km.h:703
#3  signal_pending (info=0x7ffe156fc570, vcpu=0x0) at km_signal.c:357
#4  km_post_signal (vcpu=vcpu@entry=0x0, info=info@entry=0x7ffe156fc570) at km_signal.c:434
#5  0x000000000043738f in km_signal_passthru (signo=<optimized out>, sinfo=<optimized out>, ucontext=<optimized out>) at km_vcpu_run.c:881
#6  <signal handler called>
#7  0x00000000004333d7 in km_wakeup_suspended_thread (info=info@entry=0x7ffe156fcbf0) at km_signal.c:390
#8  0x000000000043366c in km_post_signal (vcpu=vcpu@entry=0x0, info=info@entry=0x7ffe156fcbf0) at km_signal.c:439
#9  0x000000000043738f in km_signal_passthru (signo=<optimized out>, sinfo=<optimized out>, ucontext=<optimized out>) at km_vcpu_run.c:881
#10 <signal handler called>
#11 0x00000000004c66ac in read ()
#12 0x00000000004c96e2 in eventfd_read ()
#13 0x0000000000401a04 in km_wait_on_eventfd (fd=733) at /data/fedora35/km/km/km.h:375
#14 main (argc=<optimized out>, argv=<optimized out>) at km_main.c:751
(gdb) thread 2
[Switching to thread 2 (Thread 0x7f55480e4640 (LWP 576487))]
#0  0x00000000004f3cfa in __futex_abstimed_wait_common ()
(gdb) bt
#0  0x00000000004f3cfa in __futex_abstimed_wait_common ()
#1  0x000000000046b4e9 in pthread_cond_wait ()
[...]

PR #1616, opened 2022-06-24
Fix signal delivery, stopping vcpus, tighten vcpu related locks

gnode1:
Some of the tests were flaky, particularly the glibc version of them. They occasionally would fail with km locked up in processing of exit_grp. To make CI more reproducible we made these run in a loop for a few times, failing only if all attempts fail. This removes the loops, and fixes the flakiness

The fixes are:
- Two types of sending signal to VCPU - one for payload signal delivery, one for stopping VCPU. Use signal si_value field. It is 8 byte field, so we use 2 bytes for vcpu_id (instead of pointer), and another 2 bytes for flag. If flag is set this is VCPU stopping request.
- Tighten the locks on vcpus for processing vcpu pause and vcpu stop
- Drop pthread_cancel as it requires cleanup
- use pthread_cond_signal() to wake up parked VCPUs

Fixes https://github.com/kontainapp/km/issues/1455?

### DIFF
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -77,8 +77,14 @@ void km_signal_machine_fini(void)
  * Release resources allocated in km_vcpu_init(). Used in km_machine_fini() for final cleanup.
  * For normal thread completion use km_vcpu_put()
  */
-void km_vcpu_fini(km_vcpu_t* vcpu, int join_thr)
+void km_vcpu_fini(km_vcpu_t* vcpu)
 {
+   if (vcpu->vcpu_thread != 0) {
+      km_cond_signal(&vcpu->thr_cv);
+      if (pthread_join(vcpu->vcpu_thread, NULL) != 0) {
+         km_err(1, "Cannot join vcpu-%d thread", vcpu->vcpu_id);
+      }
+   }
    if (vcpu->cpu_run != NULL) {
       if (munmap(vcpu->cpu_run, machine.vm_run_size) != 0) {
          km_err(3, "munmap cpu_run for vcpu fd");
@@ -88,10 +94,6 @@ void km_vcpu_fini(km_vcpu_t* vcpu, int join_thr)
       if (close(vcpu->kvm_vcpu_fd) != 0) {
          km_err(3, "closing vcpu fd");
       }
-   }
-   if (join_thr) {
-      pthread_cancel(vcpu->vcpu_thread);
-      pthread_join(vcpu->vcpu_thread, NULL);
    }
    machine.vm_vcpus[vcpu->vcpu_id] = NULL;
    free(vcpu);
@@ -108,7 +110,7 @@ void km_machine_fini(void)
       if ((vcpu = machine.vm_vcpus[i]) == NULL) {
          break;
       }
-      km_vcpu_fini(vcpu, 1);
+      km_vcpu_fini(vcpu);
    }
    free(machine.auxv);
    if (km_guest.km_filename != NULL) {
@@ -283,11 +285,14 @@ km_vcpu_t* km_vcpu_get(void)
 
    km_mutex_lock(&machine.vm_vcpu_mtx);
    if ((vcpu = SLIST_FIRST(&machine.vm_idle_vcpus.head)) != 0) {
-      km_assert(vcpu->state == PARKED_IDLE);
       SLIST_REMOVE_HEAD(&machine.vm_idle_vcpus.head, next_idle);
       machine.vm_vcpu_run_cnt++;
+      km_lock_vcpu_thr(vcpu);
+      km_assert(vcpu->state == PARKED_IDLE);
       vcpu->state = STARTING;
+      km_unlock_vcpu_thr(vcpu);
       km_mutex_unlock(&machine.vm_vcpu_mtx);
+
       km_gdb_vcpu_state_init(vcpu);
       km_vmdriver_vcpu_init(vcpu);
       return vcpu;
@@ -303,12 +308,15 @@ km_vcpu_t* km_vcpu_get(void)
    if (km_vcpu_init(vcpu) != 0) {
       km_mutex_unlock(&machine.vm_vcpu_mtx);
       km_warnx("VCPU init failed");
-      km_vcpu_fini(vcpu, 0);
+      km_vcpu_fini(vcpu);
       return NULL;
    }
    machine.vm_vcpu_cnt++;
    machine.vm_vcpu_run_cnt++;
+   km_lock_vcpu_thr(vcpu);
    vcpu->state = STARTING;
+   km_unlock_vcpu_thr(vcpu);
+
    km_gdb_vcpu_state_init(vcpu);
    machine.vm_vcpus[vcpu->vcpu_id] = vcpu;
    km_mutex_unlock(&machine.vm_vcpu_mtx);
@@ -337,7 +345,7 @@ km_vcpu_t* km_vcpu_restore(int tid)
    vcpu->vcpu_id = slot;
    if (km_vcpu_init(vcpu) < 0) {
       km_warnx("km_vcpu_init failed - cannot restore snapshot");
-      km_vcpu_fini(vcpu, 0);
+      km_vcpu_fini(vcpu);
       return NULL;
    }
    km_gdb_vcpu_state_init(vcpu);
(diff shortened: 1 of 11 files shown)


km/km_fork.c
```diff
@@ -163,7 +163,8 @@ static void km_fork_remove_parent_vmstate(void)
    // Should try to free all of the stacks for the now defunt vcpu threads?
    for (int i = 0; i < machine.vm_vcpu_cnt; i++) {
       if (machine.vm_vcpus[i] != NULL) {
-         km_vcpu_fini(machine.vm_vcpus[i], 0);
+         machine.vm_vcpus[i]->vcpu_thread = 0;   // that thread was in parent
```
paulpopelka, line 166: Why do you do this?
km_vcpu_fini() sets machine.vm_vcpus[i] to NULL so vcpu_thread will not even exist shortly.

km/km_cpu_init.c
```diff
@@ -385,16 +394,22 @@ km_vcpu_t* km_vcpu_fetch_by_tid(pid_t tid)
    return NULL;
 }
 
+static inline void km_signal_vcpu_stop(km_vcpu_t* vcpu)
+{
+   sigval_t val = {.sival_int = vcpu->vcpu_id | 0x10000};
```
johnmuth81, line 399: Maybe something symbolic instead of hardcoded 0x10000?

johnmuth81 (commented):
Cool. One nit, but looks good otherwise.

johnmuth81 (approved):
oops, forgot to approve.

km/km_cpu_init.c
gnode1: I thought about it. Better yet, structure with a short and an enum. But then its needs to be in some shared .h file and so forth ... I thought the function name is symbolic enough.

paulpopelka, line 399: Please create a symbol for 0x10000.
It helps people know that different parts of the code are referring to the same bit.

km/km_fork.c
paulpopelka: I see what is going on.  km_vcpu_fini() will try to signal an invalid threadid when we are in the child process of a fork.
bleh.

km/km_init_guest.c
```diff
@@ -321,10 +319,14 @@ int km_clone(km_vcpu_t* vcpu,
             new_vcpu->vcpu_id,
             new_vcpu->regs.rip,
             new_vcpu->regs.rsp);
+
+   km_lock_vcpu_thr(new_vcpu);
    if (km_run_vcpu_thread(new_vcpu) < 0) {
+      km_unlock_vcpu_thr(new_vcpu);
       km_vcpu_put(new_vcpu);
```
paulpopelka, line 326: Why is it ok to not hold:

km_mutex_lock(&machine.vm_vcpu_mtx);

when you call km_vcpu_put() here?
It is held when km_vcpu_put() is called elsewhere.

gnode1: This is cleanup in case of failure to start `new_vcpu` thread. Nobody is going to look at it because of its state. I guess we can take the lock, it won't hurt. Same in `km_vcpu_fini()` - there is only one thread by then. I can move the mutex manipulation back inside the function I guess.

km/km_cpu_init.c
```diff
@@ -304,12 +306,15 @@ km_vcpu_t* km_vcpu_get(void)
    if (km_vcpu_init(vcpu) != 0) {
       km_mutex_unlock(&machine.vm_vcpu_mtx);
       km_warnx("VCPU init failed");
-      km_vcpu_fini(vcpu, 0);
+      km_vcpu_fini(vcpu);
       return NULL;
    }
    machine.vm_vcpu_cnt++;
    machine.vm_vcpu_run_cnt++;
+   km_lock_vcpu_thr(vcpu);
```
sv641, line 314: Entry and exit to free list is protected by vm_vcpu_mtx why do we need additional lock here,
If we want to protect state transitions on state then do we need to protect all changes to state variable?
Is there a implied  ownership of state variable when vcpu is changing it's own state.

gnode1: Yes, this lock protects this vcpu state transitions.

km/km_vcpu_run.c
```diff
@@ -438,6 +438,15 @@ static int hypercall(km_vcpu_t* vcpu, int* hc_ret)
    km_hc_ret_t ret = HC_CONTINUE;
    if (km_hcalls_table[hc] != NULL) {
       ret = km_hcalls_table[hc](vcpu, hc, ga_kma);
+      // check for interrupted hypercall, set restart if needed
+      if (ga_kma->hc_ret == -EINTR && vcpu->state == HCALL_INT) {
```
paulpopelka, line 442: Lock the thread state mutex here while we decide to change state?

paulpopelka (commented):
Can the pthread_setcanceltype() call in km_vcpu_run() be removed?
```
void* km_vcpu_run(km_vcpu_t* vcpu)
{
   int hc_ret = 0;
   pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);
```
We are inconsistent in our locking when making decisions based on vcpu state.
I know why it is ok sometimes.
The problems come in when code that is using an exception and doesn't lock, gets used in a different way where the locking is needed.
I'm not suggesting that this be fixed.

km/km_signal.c
```diff
@@ -359,12 +359,19 @@ int km_wakeup_suspended_thread(siginfo_t* info)
...
+{
+   sigval_t val = {.sival_int = vcpu->vcpu_id};
+
+   km_pkill(vcpu->vcpu_thread, KM_SIGVCPUSTOP, val);
+}
+
 static int km_can_interrupt(km_vcpu_t* vcpu, void* data)
 {
    int signo = *(int*)data;
    if (signo >= 0 && km_sigismember(&vcpu->sigmask, signo) == 0 && vcpu->state == HYPERCALL &&
```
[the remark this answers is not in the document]
gnode1: The lock is help by `km_vcpu_apply_all()` in this case.

km/km_cpu_init.c
```diff
@@ -77,8 +77,12 @@ void km_signal_machine_fini(void)
  * Release resources allocated in km_vcpu_init(). Used in km_machine_fini() for final cleanup.
  * For normal thread completion use km_vcpu_put()
  */
-void km_vcpu_fini(km_vcpu_t* vcpu, int join_thr)
+void km_vcpu_fini(km_vcpu_t* vcpu)
 {
+   if (vcpu->vcpu_thread != 0) {
+      km_cond_signal(&vcpu->thr_cv);
+      pthread_join(vcpu->vcpu_thread, NULL);
```
paulpopelka, line 84: If you are so inclined, verify that pthread_join() succeeded.

km/km_init_guest.c
paulpopelka: If we have 2 or more threads creating threads concurrently it seems like we could run into situations where they are both doing things with the machine.vm_idle_vcpus.head concurrently, especially if thread creation is failing for some global reason affecting all thread creators.

If something that bad is happening, this little serialization faux pas is kind of irrelevant, the entire system will be a pile of dodo shortly.

km/km_vcpu_run.c
paulpopelka: Disregard.  Not sure what I was thinking here.

km/km_cpu_init.c
paulpopelka: I don't think holding the vcpu_thr lock here does much.
Until we insert the vcpu into machine.vm_vcpus[] nothing will be able to access the vcpu except this function.

paulpopelka (approved):
A few minor comments that probably don't matter much.

km/km_signal.c
```diff
@@ -384,7 +393,19 @@ static int km_can_interrupt(km_vcpu_t* vcpu, void* data)
  */
 static void km_interrupt_thread(int signo)
 {
-   km_vcpu_apply_all(km_can_interrupt, (void*)&signo);
+   km_vcpu_t* vcpu;
+
+   if (signo >= 0) {
+      km_signal_lock();
+      if ((vcpu = TAILQ_FIRST(&km_signal_sleep_queue)) != NULL) {
+         TAILQ_REMOVE(&km_signal_sleep_queue, vcpu, signal_link);
```
paulpopelka, line 401: The vcpu is removed from the sleep queue here, then the vcpu blocked in the system call is interrupted.  The system call returns and the vcpu is removed from the sleep queue again.

I don't think TAILQ_REMOVE() will handle this correctly.

Am I misunderstanding this?

gnode1: And I think this needs to be under the if below ...

paulpopelka (approved):

PR #1634, opened 2022-07-12
initialize page tables only on kvm

sv641:
KM page tables are not required when virtualization device is kkm.

### DIFF
--- a/km/km_mem.c
+++ b/km/km_mem.c
@@ -54,6 +54,11 @@
  *
  * When using kkm driver
  * - fixed 512GB of memory is supported.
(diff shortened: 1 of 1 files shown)


sv641:
removed idt and gdt initialization when using KKM.

paulpopelka (commented):
Could you add a brief description of why the page tables are not needed when kkm is used?

sv641:
> Could you add a brief description of why the page tables are not needed when kkm is used?

done.

gnode1 (approved):

PR #1644, opened 2022-07-23
WIP: Change the snapshot mgmt interface to support additional arguments

paulpopelka:
We now send a request to the km mgmt thread with the arguments for the
snapshot request.  Previously establishing a connection was enough to
trigger a snapshot.

Also changed km_cli to search for candidate processes to snapshot based
on a list of process ids and/or a list of command names.

The bats tests pass but more bats tests need to be written for the
additional new arguments to km_cli.

### DIFF
--- a/km/km_main.c
+++ b/km/km_main.c
@@ -496,6 +496,8 @@ km_parse_args(int argc, char* argv[], int* argc_p, char** argv_p[], int* envc_p,
       char* mgt_e = getenv(KM_MGTPIPE);
       if (mgt_e != NULL) {
          mgtpipe = strdup(mgt_e);
+	 // We don't want the mgmt pipe to appear in the snapshot?
+	 unsetenv(KM_MGTPIPE);
       }
    }
 
--- a/lib/libkontain/libkontain_mgmt.h
+++ b/lib/libkontain/libkontain_mgmt.h
@@ -1,0 +1,26 @@
+#ifndef __LIBKONTAIN_MGMT_H__
+#define __LIBKONTAIN_MGMT_H__
+
+#define KM_MGMT_REQ_SNAPSHOT	1234
+
+/*
+ * Send this structure in the unix socket to the km management thread
+ * when trying to create a payload snapshot.
+ */
+#define SNAPLABELMAX 256
+#define SNAPDESCMAX 256
+#define SNAPPATHMAX 1024
+struct snapshot_req {
+   int opcode;				// what mgmt request is this.
+   int length;				// length of this request including opcode and the length
+   char label[SNAPLABELMAX];		// a label placed into the snapshot (coredump)
+   char description[SNAPDESCMAX];	// a description placed in the snapshot (coredump)
+   int live;				// if non-zero, the payload keeps running after the snapshot, if zero payload terminates
+   char snapshot_path[SNAPPATHMAX];	// path to where the snapshot should be placed.
+};
+
+struct mgmtreply {
+   int request_status;			// 0 = success, non-zero is a unix errno
+};
+
+#endif // !defined(__LIBKONTAIN_MGMT_H__)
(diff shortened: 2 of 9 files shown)


km/km_main.c
```diff
@@ -496,6 +496,8 @@ km_parse_args(int argc, char* argv[], int* argc_p, char** argv_p[], int* envc_p,
       char* mgt_e = getenv(KM_MGTPIPE);
       if (mgt_e != NULL) {
          mgtpipe = strdup(mgt_e);
+         // We don't want the mgmt pipe to appear in the snapshot?
```
gnode1, line 499: Yes you right, we don't. Drop `?`

lib/libkontain/libkontain_mgmt.h
```diff
@@ -0,0 +1,26 @@
...
+#define KM_MGMT_REQ_SNAPSHOT	1234
+
+/*
+ * Send this structure in the unix socket to the km management thread
+ * when trying to create a payload snapshot.
+ */
+#define SNAPLABELMAX 256
+#define SNAPDESCMAX 256
+#define SNAPPATHMAX 1024
+struct snapshot_req {
```
gnode1, line 13: typedef as we do everywhere else?

lib/libkontain/libkontain_mgmt.h
```diff
@@ -0,0 +1,26 @@
...
+   int length;				// length of this request including opcode and the length
+   char label[SNAPLABELMAX];		// a label placed into the snapshot (coredump)
+   char description[SNAPDESCMAX];	// a description placed in the snapshot (coredump)
+   int live;				// if non-zero, the payload keeps running after the snapshot, if zero payload terminates
+   char snapshot_path[SNAPPATHMAX];	// path to where the snapshot should be placed.
+};
+
+struct mgmtreply {
+   int request_status;			// 0 = success, non-zero is a unix errno
+};
```
gnode1, line 24: typedef?

lib/libkontain/libkontain_mgmt.h
```diff
@@ -0,0 +1,26 @@
+#ifndef __LIBKONTAIN_MGMT_H__
+#define __LIBKONTAIN_MGMT_H__
+
+#define KM_MGMT_REQ_SNAPSHOT	1234
```
gnode1, line 4: typedef enum?

lib/libkontain/libkontain_mgmt.h
```diff
@@ -0,0 +1,26 @@
...
+/*
+ * Send this structure in the unix socket to the km management thread
+ * when trying to create a payload snapshot.
+ */
+#define SNAPLABELMAX 256
+#define SNAPDESCMAX 256
+#define SNAPPATHMAX 1024
+struct snapshot_req {
+   int opcode;				// what mgmt request is this.
+   int length;				// length of this request including opcode and the length
```
gnode1, line 15: Not sure why this is needed - the struct size is fixed, so `sizeof()` will do.

lib/libkontain/libkontain_mgmt.h
```diff
@@ -0,0 +1,26 @@
...
+   char label[SNAPLABELMAX];		// a label placed into the snapshot (coredump)
+   char description[SNAPDESCMAX];	// a description placed in the snapshot (coredump)
+   int live;				// if non-zero, the payload keeps running after the snapshot, if zero payload terminates
```
gnode1, line 18: Make this into separate opcode?

gnode1 (commented):
Mostly OCD, and some questions

lib/libkontain/libkontain_mgmt.h
paulpopelka: I assume there will be other mgmt operations we want to support in the future.
If/when that happens then I would become less slothful and turn this into what it should be:
struct mgmt_request {
   int opcode;
   int length;
   union {
      struct snapshot_request {
         blah blah ...
      } snapshot;
      struct halt_and_catch_file_request {
        oh no oh no ....
      } hcf_req;
      .....
   } request_types;
};

lib/libkontain/libkontain_mgmt.h
paulpopelka: I'll change this to what it should have been.

gnode1 (approved):
I'm a bit uneasy about style in km_cli but it will do. I have a feeling this is just an initia iteration and a lot of that will be redesigned anyways.

paulpopelka:
> I'm a bit uneasy about style in km_cli but it will do. I have a feeling this is just an initia iteration and a lot of that will be redesigned anyways.

Looks like clang-format is not run on km_cli.   I'll fix the style issues.

PR #1649, opened 2022-08-01
Change km_cli -s to use the snapdir argument

paulpopelka:
Added the -t flag to km_cli to specify whether the payload terminates
after the snapshot is complete.

Fixed a bug where the request length was not supplied with km mgmt requests.

### DIFF
--- a/km_cli/km_client.c
+++ b/km_cli/km_client.c
@@ -17,7 +17,7 @@
 /*
  * Command line description:
  *
- * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l]
+ * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l] [-t]
  *
  * There are 2 parts to this command, selection of processes to snapshot and then
  * snapshotting the selected processes.
@@ -39,6 +39,7 @@
  * --snapshot=filename flag.
  *
  * The -l flag causes debug logging to stderr to happen.
+ * The -t flag causes the km payload to terminate after the snapshot is taken.
  */
 
 /*
@@ -88,6 +89,7 @@ char* cmdname;
 char* socket_name = NULL;
 
 int debug = 0;
+int terminate_app = 0;
 
 // Upper limit of -c and -p arguments
 #define MAXPIDS 32    // -p limit
@@ -102,7 +104,7 @@ void usage(void)
 {
    fprintf(stderr,
            "Usage: %s [-l] [-c commandname] [-d snapshot_dirname] [-p processid] [-s "
-           "socket_name]\n",
+           "socket_name] [-t]\n",
            cmdname);
    fprintf(stderr, "       -l   = turn on debug logging\n");
    fprintf(stderr,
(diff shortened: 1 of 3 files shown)


gnode1 (approved):
The change looks good, however I don't know if continue after snapshot was ever tested in any meaningful depth

paulpopelka:
This is what I used to test this:
```
#!/bin/bash

mkdir -p snapdir
chmod 777 snapdir
rm -f snapdir/mgmtpipe
docker run --runtime=krun --env KM_MGTPIPE=/snapdir/mgmtpipe --env KM_VERBOSE="" --mount type=bind,source=`pwd`/snapdir,target=/snapdir httpdnormalimage:latest

# To take a snapshot
# Remember km inside the container uses the path supplied with the -d flag
# and km_cli outside the container uses the path supplied with -s to find the unix socket pipe
# km created in the bind mounted snapdir directory.
km_cli -s snapdir/mgmtpipe -d /snapdir
```
The most important part is the --mount option and its argument on the "docker run" command line.

paulpopelka:
> The change looks good, however I don't know if continue after snapshot was ever tested in any meaningful depth

I'e been using "continue after snapshot" for quite a while now and it seems fine.
Well, at least the payload keeps running and the snapshots produced seem to work.

If there is something that you secretly worry about, I can try to test it.

gnode1:
Nothing in particular as far as "secretly worry about" ;-). All I'm saying
is that we have no tests for that scenario, all snapshot related tests are
destructive. I imagine something similar to the fute_snapshot test that
continues, and does it many times in a loop.

paulpopelka:
I've changed the new continue_after_snapshot test to wait 8 seconds for the hello_html_test program to startup in the background and begin listening for connections.  3 instances of that test have timed out waiting 8 seconds for the program to get going.
8 seconds seems like plenty of time.
Unless someone thinks we should give it more time, I'm not going to bother.

tests/hello_html_test.c
```diff
@@ -104,19 +104,27 @@ int main(int argc, char const* argv[])
...
-   sd = tcp_accept(listen_sd);
-   ret = read(sd, buf, 4096);
-   ret = write(sd, wbuf, sizeof(wbuf));
-   // suppress compiler warning about unused var
-   if (ret)
-      ;
-   tcp_close(sd);
+   do {
+      sd = tcp_accept(listen_sd);
+      ret = read(sd, buf, 4096);
```
gnode1, line 118: `sizeof(buf)` ?

km_cli/km_client.c
```diff
@@ -17,7 +17,7 @@
 /*
  * Command line description:
  *
- * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l]
+ * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l] [-t]
```
gnode1, line 20: This changes the default behavior. I'd rather we keep it. IOW have an option to continue instead.

gnode1 (approved):

km_cli/km_client.c
paulpopelka: Only the -s flag implied the payload terminated after the snapshot was completed.
Nothing used km_cli until I started testing it in the bats basic_snapshot test a few weeks ago.
That testing used the new flags I added, nothing used the existing -s flag.
The new flags all assumed that snapshot requests did not terminate the payload.
This PR is the first time we test with the -s flag, so nothing will be affected by this small change in semantics.
I will leave things as they are now in this change.

tests/hello_html_test.c
paulpopelka: Will fix.

PR #1651, opened 2022-08-09
change variable signedness so we can detect read/write errors

paulpopelka:
Plus add checks for ignored errors in the hello_html_test program.

Decided to fix these "problems" since hello_html_test was involved in the odd km failure where the kernel returned an fd that km thought was already in use.


gnode1 (approved):
couple of comments

PR #1654, opened 2022-08-10
wait for thread to start instead of sleep(2)

gnode1:


paulpopelka (approved):
I think this change will help.  Do we want to put an upper limit on how long we are willing to sleep in the loop.  I'm being paranoid. 

One comment though:
Can we change the check for the value returned by pthread_cancel() from this:
ASSERT(s == 0 || s == ESRCH);
to this:
ASSERT(s == 0);
?

gnode1:
> I think this change will help. Do we want to put an upper limit on how long we are willing to sleep in the loop. I'm being paranoid.
> 

pthread_create succeed so thread should start
 Worse comes to worse bats will timeout

> One comment though: Can we change the check for the value returned by pthread_cancel() from this: ASSERT(s == 0 || s == ESRCH); to this: ASSERT(s == 0); ?

I thought I did.

paulpopelka:
> > One comment though: Can we change the check for the value returned by pthread_cancel() from this: ASSERT(s == 0 || s == ESRCH); to this: ASSERT(s == 0); ?
> 
> I thought I did.

I don't see that change there.

paulpopelka (approved):

PR #1656, opened 2022-08-13
Change wait in snapshot test from process exists to mgt pipe exists

johnmuth81:
Fixes recent CI failures in snapshot tests

### DIFF
--- a/tests/km_core_tests.bats
+++ b/tests/km_core_tests.bats
@@ -1099,9 +1099,8 @@ fi
    mkdir -p $SNAPDIR
    rm -f $MGMTPIPE
    KM_MGTPIPE=$MGMTPIPE km_with_timeout hello_html_test$ext $snapshot_test_port &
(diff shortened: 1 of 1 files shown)


tests/km_core_tests.bats
```diff
@@ -1101,7 +1101,8 @@ fi
    KM_MGTPIPE=$MGMTPIPE km_with_timeout hello_html_test$ext $snapshot_test_port &
    # Wait for the process to exist
```
paulpopelka, line 1102: Might want to change the comment to say process and mgmt pipe to exist.

paulpopelka (approved):
Good, one instance of pidof removed.
Thanks for finding and fixing this.

tests/km_core_tests.bats
johnmuth81: good point. dead code needed to be removed too.

PR #1657, opened 2022-08-15
remove dead code in decode + don't be silent about unrecognized opcodes

johnmuth81:


gnode1 (approved):
LGTM - with one question: Does TODO on line 728 still apply? (Generalize to all instructions with ModR/M byte)

johnmuth81:
> LGTM - with one question: Does TODO on line 728 still apply? (Generalize to all instructions with ModR/M byte)

sadly, yes.

PR #1659, opened 2022-08-16
generalize 1 byte indirect decode

johnmuth81:


gnode1 (approved):
Do you want to add tests for the new logic?

PR #1665, opened 2022-08-24
eliminate race between exit_grp and clone

gnode1:
If `km_vcpu_get()` is called after `exit_grp()` it is possible to get a vcpu that already done `pthread_exit()`. That vcpu will never run hence never exit hence never accounted as exit. km main thread would wait forever.

### DIFF
--- a/km/km_cpu_init.c
+++ b/km/km_cpu_init.c
@@ -283,6 +283,9 @@ km_vcpu_t* km_vcpu_get(void)
 {
    km_vcpu_t* vcpu;
 
+   if (machine.exit_group != 0) {
+      return NULL;
+   }
    km_mutex_lock(&machine.vm_vcpu_mtx);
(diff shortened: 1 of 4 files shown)


km/km_cpu_init.c
```diff
@@ -283,6 +283,9 @@ km_vcpu_t* km_vcpu_get(void)
 {
    km_vcpu_t* vcpu;
 
+   if (machine.exit_group != 0) {
```
sv641, line 286: Consider current thread crossing this line and getting scheduled out other thread then runs exit group  and exit_grp run's farther before first thread gets scheduled back.

gnode1: You right.

sv641 (approved):

sv641 (approved):

PR #1667, opened 2022-09-01
Add code to snapshot/recover eventfd's

paulpopelka:
The snapshot code would treat eventfd's as though they were epoll fd's
which caused problems if you called epoll_wait() on an eventfd.
So, added code to support snapshoting and recovering eventfd.s
Note that the snapshot code called epoll fd's event fd's so had to
rename some functions too.

Added eventfd testing to the snapshot_test in bats.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -1962,6 +1962,48 @@ static inline size_t fs_core_write_socket(char* buf, size_t length, km_file_t* f
    return cur - buf;
 }
 
+/*
+ * Open /proc/self/fdinfo/%d, read its contents, find the field named by
+ * fieldname, convert its associated value to binary and return that value.
+ * The passed fd should be an fd for an eventfd.
+ * We expect the fdinfo file to look something like this:
+ * [paulp@home km]$ sudo cat /proc/1022/fdinfo/3
+ * pos:	0
+ * flags:	02004002
+ * mnt_id:	15
+ * ino:	12162
+ * eventfd-count:                0
+ * eventfd-id: 4
+ * [paulp@home km]$
+ */
+static inline void extract_fdinfo_field(int fd, char* fieldname, uint64_t* value)
+{
+   char procfile[128];
+   char fdinfobuf[256];
+   snprintf(procfile, sizeof(procfile), "/proc/self/fdinfo/%d", fd);
+   int pfd = open(procfile, O_RDONLY);
+   if (pfd < 0) {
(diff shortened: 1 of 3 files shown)


km/km_filesys.c
```diff
@@ -1937,7 +1939,7 @@ static inline size_t fs_core_write_socket(char* buf, size_t length, km_file_t* f
    // Don't snapshot a socketpair with in flight data.
    if (file->how == KM_FILE_HOW_SOCKETPAIR0 || file->how == KM_FILE_HOW_SOCKETPAIR1) {
       if (ioctlfionread(fd) > 0) {
-         km_errx(1, "Couldn't perform snapshot, socketpair fd %d has queued data", fd);
+         km_warn("Couldn't perform snapshot, socketpair fd %d has queued data", fd);
```
paulpopelka, line 1942: We probably don't want this change.
It is needed to get the node snapshot to work.

km/km_filesys.c
```diff
@@ -1898,7 +1900,7 @@ static inline size_t fs_core_write_nonsocket(char* buf, size_t length, km_file_t
       fnote->data = file->ofd;   // default to ofd. override based on file type
       // If a non-std{in,out,err} pipe contains data we don't snapshot
       if (fd > 2 && ioctlfionread(fd) > 0) {
-         km_errx(1, "Can't take a snapshot, fifo fd %d has buffered data", fd);
+         km_warn("Can't take a snapshot, fifo fd %d has buffered data", fd);
```
paulpopelka, line 1903: We probably don't want this change.
It is needed to get the node snapshot to work.

km/km_filesys.c
```diff
@@ -1960,12 +1962,23 @@ static inline size_t fs_core_write_socket(char* buf, size_t length, km_file_t* f
...
+   size_t remain = length;
+
+   km_infox(KM_TRACE_SNAPSHOT, "fd=%d eventfd", fd);
+   cur += km_add_note_header(cur, remain, KM_NT_NAME, NT_KM_EVENTFD, sizeof(km_nt_file_t));
+   km_nt_file_t* fnote = (km_nt_file_t*)cur;
+   cur += sizeof(km_nt_file_t);
+   fnote->size = sizeof(km_nt_file_t);
+   fnote->fd = fd;
+   fnote->flags = file->flags;
+
```
paulpopelka, line 1977: We need to save the current value of the eventfd counter.

gnode1 (approved):
The eventfd related changes LGTM. Other changes need to be reconciled before the merge

PR #1668, opened 2022-09-02
Serge/handle dup

gnode1:
Record dev/inode numbers for file descriptor on snapshot. Use that to discover duped fds on restore. Simply do dup{2,3} on restore when dup is detected.

Added tests for duped files and sockets.

This implementation has a drawback. It will incorrectly handle two fds that were open on the same file, not duped. The file offset will by tied for them together after the recover as the algorithm will mistake them for dups. See https://github.com/kontainapp/km/issues/1671

### DIFF
--- a/km/km_coredump.h
+++ b/km/km_coredump.h
@@ -117,6 +117,8 @@ typedef struct km_nt_guest {
 typedef struct km_nt_file {
    Elf64_Word size;    // Size of record
    Elf64_Word fd;      // Open fd number
+   Elf64_Word dev;     // Device number - for dup detection
+   Elf64_Word ino;     // Inode number
    Elf64_Word how;     // How file was created
    Elf64_Word flags;   // open(2) flags
    Elf64_Word mode;    // file mode (includes type)
@@ -133,10 +135,14 @@ typedef struct km_nt_file {
    // Followed by data buffered in a pipe
 } km_nt_file_t;
 #define NT_KM_FILE 0x4b4d4644   // "KMFD" no null term
+// eventfd
+#define NT_KM_EVENTFD 0x4b4d4556   // "KMEV" no null term
 
 typedef struct km_nt_socket {
    Elf64_Word size;      // Size of record
    Elf64_Word fd;        // Open fd number
+   Elf64_Word dev;       // Device number - for dup detection
+   Elf64_Word ino;       // Inode number
    Elf64_Word how;       // How socket was created
    Elf64_Word state;     // state of socket
    Elf64_Word backlog;   // listen backlog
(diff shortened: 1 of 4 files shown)


km/km_filesys.c
```diff
@@ -2975,7 +3043,9 @@ static int km_fs_recover_open_socket(char* ptr, size_t length)
          return -1;
       }
       if (bind(hostfd, sa, nt_sock->addrlen) < 0) {
-         km_warn("recover bind failed");
+         km_warn("recover bind failed %s:%d",
```
paulpopelka, line 3046: This assumes address family is AF_INET.
Maybe an assert to let us know when we are being feed some other address family.

km/km_filesys.c
```diff
@@ -2139,6 +2169,33 @@ static inline void km_fs_recover_fd(int guestfd, int hostfd, int flags, char* na
             how);
 }
 
+typedef struct km_dup_data {
+   dev_t dev;
+   ino_t ino;
+} km_dup_data_t;
+
+km_dup_data_t* dup_data;
```
paulpopelka, line 2177: declare static?

km/km_filesys.c
```diff
@@ -3085,5 +3168,6 @@ int km_fs_recover(char* notebuf, size_t notesize)
    if (km_snapshot_notes_apply(notebuf, notesize, NT_KM_EPOLLFD, km_fs_recover_epollfd) < 0) {
       km_errx(2, "recover open epollfd's failed");
    }
+   free(dup_data);
```
paulpopelka, line 3171: set dup_data to null after the free.

paulpopelka (approved):
Looks good to me.
A few minor comments.
I suppose I need to get my change checked in.
There are few things that need to be fixed in my change before it goes in.

km/km_coredump.h
```diff
@@ -172,18 +178,18 @@ typedef struct km_nt_event {
...
+typedef struct km_nt_epollfd {
    Elf64_Word size;         // Size of record
    Elf64_Word fd;           // Open event fd
+   Elf64_Word dev;          // Device number - for dup detection
+   Elf64_Word ino;          // Inode number
    Elf64_Word flags;        // flags
    Elf64_Word event_size;   // size of event records that follow
    Elf64_Word nevent;       // number of event records that follow
-} km_nt_eventfd_t;
+   // Followed by event records km_nt_event_t
```
sv641, line 189: add **km_nt_event_t nt_event[0];**

sv641: makes code easier to write.

km/km_filesys.c
```diff
@@ -2006,6 +2015,11 @@ static inline void extract_fdinfo_field(int fd, char* fieldname, uint64_t* value
 #define EVENTFD_COUNT "eventfd-count:"
 static inline size_t fs_core_write_eventfd(char* buf, size_t length, km_file_t* file, int fd)
 {
+   struct stat st = {};
+   if (fstat(fd, &st) < 0) {
+      km_warn("Can't take a fstat failed fd=%d", fd);   // TODO: return error
```
sv641, line 2020: same as above

sv641 (approved):

km/km_coredump.h
gnode1: I ws thinking about it but decided against it for this PR. There are many code improvements like doing `typedef enum` instead of sets of #define and such. Another PR for that

PR #1670, opened 2022-09-03
Add support for snapshoting data buffered in a socketpair

paulpopelka:
Also support for snapshotting data buffered in a pipe.
The buffered data is stored in the elf note for the write side
of the pipe or socketpair fd's in the snapshot.
The data in the pipe or socketpair is read out to perform the
snapshot and written back into the write side of the pipe so
the application can get its data if it were to resume after the
snapshot request returns.
Changed the bats snapshot_fail_test to now verify that data buffered
in a pipe or socketpair is now available when the snapshot is
restarted.

### DIFF
--- a/km/km_coredump.h
+++ b/km/km_coredump.h
@@ -128,7 +128,9 @@ typedef struct km_nt_file {
     *   __S_ISOCK  - 'other' fd in socketpair
     */
    Elf64_Off data;   // depends on file type.
+   Elf64_Word datalength; // bytes of pipe contents following filename
    // Followed by file name
+   // Followed by pipe or socketpair contents
 } km_nt_file_t;
 #define NT_KM_FILE 0x4b4d4644   // "KMFD" no null term
 
@@ -143,7 +145,9 @@ typedef struct km_nt_socket {
    Elf64_Word protocol;
    Elf64_Word other;   // 'other' fd for socketpair(2)
    Elf64_Word addrlen;
+   Elf64_Word datalength; // bytes of to write back into this side
    // Address follows
+   // Data queued in socket follows the address
 } km_nt_socket_t;
 #define NT_KM_SOCKET 0x4b4d534b   // "KMSK" no null term
 
@@ -159,9 +163,15 @@ typedef struct km_nt_socket {
 #define KM_NT_SKSTATE_CONNECT 4
 #define KM_NT_SKSTATE_ERROR 5
 
+// Use a function so that we consistently roundup note related pieces in the rest of the code.
+static inline size_t km_nt_chunk_roundup(size_t size)
+{
+   return roundup(size, 4);
+}
+
 static inline size_t km_nt_file_padded_size(const char* str)
 {
-   return roundup(strlen(str) + 1, 4);
+   return km_nt_chunk_roundup(strlen(str) + 1);
 }
 
 // Single event on eventfd (epoll_create)
--- a/km/km_coredump.h
+++ b/km/km_coredump.h
@@ -127,8 +127,8 @@ typedef struct km_nt_file {
     *   __S_IFIFO  - 'other' fd in pipe
     *   __S_ISOCK  - 'other' fd in socketpair
     */
-   Elf64_Off data;   // depends on file type.
-   Elf64_Word datalength; // bytes of pipe contents following filename
+   Elf64_Off data;          // depends on file type.
+   Elf64_Word datalength;   // bytes of pipe contents following filename
    // Followed by file name
    // Followed by pipe or socketpair contents
 } km_nt_file_t;
@@ -145,7 +145,7 @@ typedef struct km_nt_socket {
    Elf64_Word protocol;
    Elf64_Word other;   // 'other' fd for socketpair(2)
    Elf64_Word addrlen;
-   Elf64_Word datalength; // bytes of to write back into this side
+   Elf64_Word datalength;   // bytes of to write back into this side
    // Address follows
    // Data queued in socket follows the address
 } km_nt_socket_t;
--- a/km/km_coredump.h
+++ b/km/km_coredump.h
@@ -130,7 +130,7 @@ typedef struct km_nt_file {
    Elf64_Off data;          // depends on file type.
    Elf64_Word datalength;   // bytes of pipe contents following filename
    // Followed by file name
-   // Followed by pipe or socketpair contents
+   // Followed by data buffered in a pipe
 } km_nt_file_t;
 #define NT_KM_FILE 0x4b4d4644   // "KMFD" no null term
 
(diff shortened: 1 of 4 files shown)


paulpopelka:
FYI:
Another solution has been suggested.
Put the implementation of pipes and socketpairs into km, in other words don't go into the kernel for reading/writing these pipe like things.
This way the data will be buffered inside of km itself and can be snapshotted from km memory instead of needing to read the pipe contents.
The concern is handling pipes where one end is outside of the process being snapshotted which the implementation in this review will not handle.
Since we don't support snapshot of multiple payloads concurrently, this is not a concern.  But could be in the future.

km/km_filesys.c
```diff
@@ -1864,32 +1872,71 @@ size_t km_fs_core_notes_length()
...
-                   km_nt_file_padded_size(file->name);
+                   km_nt_file_padded_size(file->name) + km_nt_chunk_roundup(queuedbytes);
          }
       }
    }
    return ret;
 }
 
+static inline size_t
+fs_core_save_pipe_contents(char* buf, size_t length, km_file_t* file, int writefd, size_t queuedbytes)
```
gnode1, line 1896: maybe mention it applies to sockets too?

km/km_filesys.c
```diff
@@ -1864,32 +1872,71 @@ size_t km_fs_core_notes_length()
...
+   km_infox(KM_TRACE_SNAPSHOT, "fd=%d %s, queuedbytes %ld", fd, file->name, queuedbytes);
    char* cur = buf;
    size_t remain = length;
    cur += km_add_note_header(cur,
                              remain,
                              KM_NT_NAME,
                              NT_KM_FILE,
-                             sizeof(km_nt_file_t) + km_nt_file_padded_size(file->name));
+                             sizeof(km_nt_file_t) + km_nt_file_padded_size(file->name) +
+                                 roundup(queuedbytes, 4));
```
gnode1, line 1939: `km_nt_chunk_roundup`? Also what about the `= km_nt_chunk_roundup(fnote->datalength);` down below on line 1957?

km/km_filesys.c
```diff
@@ -2201,7 +2298,11 @@ static inline int km_fs_recover_pipe(km_nt_file_t* nt_file, char* name)
       return -1;
    }
 
-   km_infox(KM_TRACE_SNAPSHOT, "recovered pipe: rfd=%d wfd=%d flags=0x%x", rfd, wfd, syscall_flags);
+   // Recover queued pipe contents
+   if (km_fs_recover_pipedata(nt_file, pipedata) != 0) {
```
gnode1, line 2302: You simply pass the return on line 2276, an do this here. Any reason?

tests/snapshot_fail_test.c
```diff
@@ -40,10 +40,11 @@ char* SOCKET_PORT = "SOCKET_PORT";
  * A simple test to verify that snapshots fail if the following are found in the
  * payload being snapshotted.
  * - epoll fd's with events
- * - pipes with queued data
- * - socketpair's with queued data
  * - sockets that are in a connected state.
  * - interval timers that are running (setitimer() and timer_create())
+ * Now the following conditions are supported.  So test that they work.
+ * - pipes with queued data
+ * - socketpair's with queued data
```
gnode1, line 47: Kind of contradicts the name of the test. No bug deal, just a bit confusing

gnode1 (approved):
LGTM with some question modtly for my understanding. Maybe a bit of extra comments

km/km_coredump.h
```diff
@@ -159,9 +163,15 @@ typedef struct km_nt_socket {
 #define KM_NT_SKSTATE_CONNECT 4
 #define KM_NT_SKSTATE_ERROR 5
 
+// Use a function so that we consistently roundup note related pieces in the rest of the code.
+static inline size_t km_nt_chunk_roundup(size_t size)
+{
+   return roundup(size, 4);
```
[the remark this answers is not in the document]
paulpopelka: I guess the fields in a note are supposed to be 4 byte aligned, but some people say they should be 8 byte aligned on a 64 bit architecture, but linux stays with 4 byte alignment on a 64 bit architecture.
So, we make sure what we add to the note will result in the following note being 4 byte aligned so that its fields will be 4 byte aligned.  The following may be helpful:
https://groups.google.com/g/generic-abi/c/vT-_QVcckXo?pli=1
It just further re-enforces the confusion for me.  So, just do what keeps them pacified.

km/km_filesys.c
```diff
@@ -1898,31 +1945,45 @@ static inline size_t fs_core_write_nonsocket(char* buf, size_t length, km_file_t
...
+      cur += km_nt_chunk_roundup(fnote->datalength);
+   }
+
    return cur - buf;
 }
 
+/*
+ * Handles sockets created by socket() and socketpair().
+ */
 static inline size_t fs_core_write_socket(char* buf, size_t length, km_file_t* file, int fd)
```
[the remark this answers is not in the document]
paulpopelka: It does seem rather large for an inline.
inline seems advisory.  The compiler can ignore the inline directive under many conditions.
I can remove "inline".

km/km_filesys.c
paulpopelka: Will fix.

km/km_filesys.c
paulpopelka: Sloth!  Will fix.

tests/snapshot_fail_test.c
paulpopelka: Yes, I mentioned this to you in our conversation about recovering "pipe" contents before this change was created.
The test should probably be renamed to snapshot_cornercase_test.

km/km_filesys.c
paulpopelka: Added a comment describing the interface this function provides.

km/km_coredump.h
```diff
@@ -127,8 +127,10 @@ typedef struct km_nt_file {
     *   __S_IFIFO  - 'other' fd in pipe
     *   __S_ISOCK  - 'other' fd in socketpair
     */
-   Elf64_Off data;   // depends on file type.
+   Elf64_Off data;          // depends on file type.
+   Elf64_Word datalength;   // bytes of pipe contents following filename
    // Followed by file name
+   // Followed by pipe or socketpair contents
```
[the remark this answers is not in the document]
paulpopelka: I changed the comment, see if it helps.

km/km_coredump.h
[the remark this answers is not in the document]
gnode1: My point was simply that 4 seems to be a magic number. I'd like to see `const int`, with that explanation next to definition.

km/km_coredump.h
```diff
@@ -130,7 +130,7 @@ typedef struct km_nt_file {
    Elf64_Off data;          // depends on file type.
    Elf64_Word datalength;   // bytes of pipe contents following filename
    // Followed by file name
-   // Followed by pipe or socketpair contents
+   // Followed by data buffered in a pipe
```
gnode1, line 133: My point was more about the comment on line 132 and this one, being confused when looked at together

gnode1 (approved):

PR #1673, opened 2022-09-08
Change bats basic_snapshot to tolerate multiple pids from pidof

paulpopelka:
The basic_snapshot test uses the pidof command to find the pid of a command it wants to snapshot.  When the bats tests run with concurrency enabled pidof can return multiple pids which isn't handled. Add a small loop to traverse the list of pids returned by pidof to find which pid we need for the test.

### DIFF
--- a/tests/km_core_tests.bats
+++ b/tests/km_core_tests.bats
@@ -1103,7 +1103,16 @@ fi
    tries=5
    while [ ! -S ${MGMTPIPE} ] && [ $tries -gt 0 ]; do sleep 1; tries=`expr $tries - 1`; done
    assert [ $tries -gt 0 ]
-   pid=`pidof hello_html_test$ext`
+   pidlist=`pidof hello_html_test$ext`
+   # Find our pid if pidof returned multiple pids
+   pid=""
+   for p in $pidlist; do
+      if lsof -p $p | grep " 729u .* $MGMTPIPE " >/dev/null 2>&1; then
+         pid=$p
+         break
+      fi
+   done
+   assert [ ! -z "$pid" ]
    run ${KM_CLI_BIN} -p $pid -d $SNAPDIR
    assert_success
    assert [ -f $SNAPDIR/hello_html_test$ext.$pid.kmsnap ]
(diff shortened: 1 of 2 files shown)


tests/km_core_tests.bats
```diff
@@ -1103,7 +1103,16 @@ fi
    tries=5
    while [ ! -S ${MGMTPIPE} ] && [ $tries -gt 0 ]; do sleep 1; tries=`expr $tries - 1`; done
    assert [ $tries -gt 0 ]
-   pid=`pidof hello_html_test$ext`
+   pidlist=`pidof hello_html_test$ext`
+   # Find our pid if pidof returned multiple pids
+   pid=""
+   for p in $pidlist; do
+      if lsof -p $p | grep " 729u .* $MGMTPIPE " >/dev/null 2>&1; then
```
gnode1, line 1110: `|` -> `|&`, `grep -q` would make redirects unnecessary

paulpopelka: I'll change that.  It occurs to me that lsof is probably not in the test container.
I will switch over to looking at /proc/$pid/fd/729
I wonder if the readlink command is available.

gnode1 (approved):

tests/km_core_tests.bats
sv641: what is 729u. 

MGTPIPE is /tmp/mgmtpipe.$$. $$ is the shell pid. Is this ok to create uniqueness?

gnode1: `lsof` is not there, you are right. We can put it there but it will require to rebuild buildenv image. 

`readlink` is there, it is part of coreutils, so it's there on most systems, even very small ones.

paulpopelka: > what is 729u.
> 
> MGTPIPE is /tmp/mgmtpipe.$$. $$ is the shell pid. Is this ok to create uniqueness?

The km management pipe is always on fd 729.

sv641 (approved):

PR #1674, opened 2022-09-08
limit inode based dup check to files and sockets only

gnode1:
The dup detection that was implemented is based on inode # and dev being the same for dups, _**and not the same for non dups**_

Well, turns out, both end of the pipe have the same inode #. Also, all of the fds in the same eventfd group have the same inode #.

Also all kvm fds other than the first `/dev/kvm` open are the same inode #. The latter doesn't affect us, just an observation.

[before.txt](https://github.com/kontainapp/km/files/9529958/before.txt)

This restrict the inode # based dup check to files and regular sockets. Simple node snapshot test passes.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -3167,13 +3167,6 @@ static int km_fs_recover_open_socket(char* ptr, size_t length)
       km_warnx("nt_km_socket_t size mismatch - old snapshot?");
       return -1;
    }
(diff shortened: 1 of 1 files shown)


sv641 (approved):
Does this mean dup detection for event and epoll has to be implemented in a different way?

gnode1:
All of this needs to be done differently. This PR is a quick patch for right now. I am working on the real solution but it takes a bit longer

johnmuth81 (approved):
Yep, that fixes what I'm seeing.

PR #1675, opened 2022-09-09
Revisited duped fd processing

gnode1:
This is better/correct way to handle duped fd.

During the run we monitor all sources of duped fds - dup{,2,3} and F_DUPFD fcntl, also we monitor close. That allows us to maintain a table of dup groups. Each group is a list of fds that are dups of each other. There could be many of them.

On snapshot we create a new note to pass these groups along. On restore we restore that before fds. When we restore fds we check with the duped groups. FOr each fd that is restored first in the group we restore in normally. But for the subsequent ones we do the dup.

tested with normal test extended a bit. Also checked node snapshot works. Will be testing more, likely adding more tests. But appreciate a review.

### DIFF
--- a/km/km_coredump.h
+++ b/km/km_coredump.h
@@ -117,8 +117,6 @@ typedef struct km_nt_guest {
 typedef struct km_nt_file {
    Elf64_Word size;    // Size of record
    Elf64_Word fd;      // Open fd number
-   Elf64_Word dev;     // Device number - for dup detection
-   Elf64_Word ino;     // Inode number
    Elf64_Word how;     // How file was created
    Elf64_Word flags;   // open(2) flags
    Elf64_Word mode;    // file mode (includes type)
@@ -141,8 +139,6 @@ typedef struct km_nt_file {
 typedef struct km_nt_socket {
    Elf64_Word size;      // Size of record
    Elf64_Word fd;        // Open fd number
-   Elf64_Word dev;       // Device number - for dup detection
-   Elf64_Word ino;       // Inode number
    Elf64_Word how;       // How socket was created
    Elf64_Word state;     // state of socket
    Elf64_Word backlog;   // listen backlog
@@ -192,7 +188,7 @@ static inline size_t km_nt_file_padded_size(const char* str)
    return km_nt_chunk_roundup(strlen(str) + 1);
 }
 
-// Single event on eventfd (epoll_create)
+// Single event on epollfd (epoll_create)
 typedef struct km_nt_event {
    Elf64_Word fd;   // fd to monitor
    Elf64_Word event;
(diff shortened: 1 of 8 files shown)


km/km_coredump.h
```diff
@@ -111,14 +111,28 @@ typedef struct km_nt_guest {
 #define NT_KM_GUEST 0x4b4d4754       // "KMGT" no null term
 #define NT_KM_DYNLINKER 0x4b4d444c   // "KMDL" no null term
 
+/*
+ * Elf note for dup data
```
paulpopelka, line 115: Could you note that these are only used in snapshots.

paulpopelka:
I'm not sure if the following condition is handled.

assume fd 0-5 are already open.
create a pipe which yields fd 6 and 7
close fd 4
dup fd 6 which yields fd 4, fd 4 has how value of "pipe" but ofd is -1 so there is no link between fd 4 and fd 7.
And, fd 7 still refers to fd 6.
close fd 6
And still fd 7's ofd refers to fd 6.

Now fd 4 and fd 7 are the ends of the pipe.
But fd 4 has an ofd value of -1 and fd 7 refers to fd 6 which is closed.

take a snapshot.

When the snapshot is restored, fd 4 is discovered first since we recover in ascending fd order (0 -> infinity).
So we try to create a pipe but fd 4's ofd doesn't refer to fd 7.  
The we discover fd 7 and create another pipe using fd 6 and 7.

I think this is happening by inspection of the code.  I should probably write a test that actually does this and see what happens.

km/km_coredump.h
gnode1: All of these here are only used in snapshots as described in the commment up above.

gnode1:
I think your analysis is spot on, this case will fail. With this in mind - should we disallow dups on pairs endpoints? Or at least disable snapshot if there were any?

johnmuth81 (approved):
lgtm

km/km_filesys.c
```diff
@@ -85,6 +88,123 @@ void km_set_file_used(km_file_t* file, int val)
...
+   group->fds[group->size++] = fd;
+}
+
+/*
+ * Called during normal run to record dup operation
+ */
+static void km_fs_add_to_dup_data(int new_fd, int old_fd)
+{
+   if (machine.mmaps.recovery_mode != 0) {
+      // during snapshot recovery dup data gets restored fist, then used to restore files
```
paulpopelka, line 136: first not fist.

km/km_filesys.c
```diff
@@ -503,6 +625,7 @@ uint64_t km_fs_fcntl(km_vcpu_t* vcpu, int fd, int cmd, uint64_t arg)
                                    (cmd == F_DUPFD) ? 0 : O_CLOEXEC,
                                    file->sockinfo,
                                    file->how);
+            km_fs_add_to_dup_data(ret, host_fd);
          } else {
             ret = km_add_guest_fd_internal(vcpu,
```
paulpopelka, line 630: We don't want to save dup info for non-socket files here but further down in km_fs_dup() we do.
Why?

paulpopelka (commented):
Looks ok.
I have one question that I'd like you to answer.

km/km_filesys.c
gnode1: I think its a bug. I guess we haven't seen real file fds passed that way

sv641 (approved):
We need a comment saying we don'r support dup  info across exec.

paulpopelka (approved):

PR #1676, opened 2022-09-13
Change snapshot to handle data buffered in half open pipes and socket…

paulpopelka:
…pairs.

Move buffered pipe and socketpair data to the elf note for the read end of the pipe or socketpair.  We formerly kept the data in the elf note for the write end of the pipe/socketpair.  With a half open pipe or socketpair the written fd is closed after all data has been written.  The data stays queued until the other end is ready to read it.
We can't really tell where the other end of these half open pipes/socketpairs is.  It may still be in another process or it may have been created local to the process and then closed.  We can snapshot the pipe/socketpair that is local to the process being snapshotted. We can't snapshot the data buffered in a pipe/socketpair where the other end is in a different process since we don't support multi-process snapshot.

So a new restriction on snapshot is being imposed.  If a process has successfully forked, it can not be snapshotted.  This avoids dealing with "apparently" half open pipe/socketpairs where the other end may be in another process.

Added code to save and restore the size of data that can be buffered in the kernel in a pipe.  No test for this yet.

Added new bats tests for snapshotting processes with half open pipes and socketpairs and processes that have forked.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -1981,7 +1981,7 @@ size_t km_fs_dup_notes_length(void)
 static int ioctlfionread(int fd)
 {
    int bytesavailable = 0;
-   if (fd >= 0 && ioctl(fd, FIONREAD, &bytesavailable) < 0) {
+   if (ioctl(fd, FIONREAD, &bytesavailable) < 0) {
       km_err(1, "ioctl FIONREAD on fd %d failed", fd);
    }
    return bytesavailable;
@@ -2008,7 +2008,7 @@ size_t km_fs_core_notes_length(void)
          } else if (file->sockinfo == NULL) {
             queuedbytes = 0;
             if (file->how == KM_FILE_HOW_PIPE_0) {
-               // We are looking at the write end of a pipe, find out how much data is queued
+               // We are looking at the read end of a pipe, find out how much data is queued
                queuedbytes = ioctlfionread(i);
             }
             ret += km_note_header_size(KM_NT_NAME) + sizeof(km_nt_file_t) +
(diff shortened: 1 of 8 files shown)


km/km_filesys.c
```diff
@@ -1951,27 +1962,55 @@ static inline void do_full_write(int fd, char* bufp, size_t bufl)
...
+            km_err(1, "Unable to recreate half open socketpair, fd %d", fd);
+         }
+      } else {
+         if (pipe2(fdpair, file->flags) < 0) {
+            km_err(1, "Unable to recreate half open pipe, fd %d", fd);
+         }
+      }
+      do_full_write(fdpair[1], buf, queuedbytes);
+      close(fdpair[1]);
+      if (dup2(fdpair[0], fd) < 0) {
```
paulpopelka, line 1990: We need to handle the case where the new pipe/socketpair gets the same fd we started with.  In that case the dup() call is pointless and the close(fdpair[0]) is not needed.

paulpopelka:
This change does not include the latest changes for handling fd's that are dups.
I just want this change to be commented on and I will bring it up to the top of master soon.

km/km_snapshot.c
```diff
@@ -520,6 +521,10 @@ int km_snapshot_create(km_vcpu_t* vcpu, char* label, char* description, char* du
       km_warnx("Cannot create snapshot with GDB running");
       return -EBUSY;
    }
+   if (km_have_forked() != 0) {
```
sv641, line 524: From what I understand this gets wiped out on exec.

paulpopelka: yes, I need to add it to the state that is saved across exec.

km/km_filesys.c
```diff
@@ -1951,27 +1962,55 @@ static inline void do_full_write(int fd, char* bufp, size_t bufl)
...
+   // Read the bytes queued in the pipe into the memory buf points to
+   do_full_read(fd, buf, queuedbytes);
 
-   // now write the bytes back into the pipe
-   do_full_write(writefd, buf, queuedbytes);
+   if (writefd < 0) {
+      // The write side of this pipe is closed.  So we need to recreate the pipe
+      // to put the data back into the pipe buffer and then close the write end
+      // after writing the data.
+      close(fd);
```
sv641, line 1976: as you mentioned. You may want to remove this line.
If nothing else it makes code easier to analyse wrt same fd being given back on pipe2/sockerpair.

sv641: (i.e) you got a fd same as this and you dup and close the same fd.

km/km_filesys.c
```diff
@@ -1951,27 +1962,55 @@ static inline void do_full_write(int fd, char* bufp, size_t bufl)
 static inline size_t
 fs_core_save_pipe_contents(char* buf, size_t length, km_file_t* file, int writefd, size_t queuedbytes)
```
sv641, line 1963: replace file and writefd paramenters with fd. This will make code more understandable.

km/km_snapshot.c
sv641: I think km_called_via_exec can be reused for this purpose.

paulpopelka: If you are running a program that does not fork but does exec to another program then no new process is created and we should not see any cases of ambiguous half open pipes/socketpairs.
I think we need to propagate the fork count in exec.

km/km_filesys.c
paulpopelka: Will fix.

km/km_filesys.c
paulpopelka: Will fix.

km/km_filesys.c
```diff
@@ -1839,6 +1839,17 @@ uint64_t km_fs_prlimit64(km_vcpu_t* vcpu,
    return ret;
 }
 
+// Given a file pointer return its associated fd number.
```
sv641, line 1842: this function is no longer used

km/km_fork.c
```diff
@@ -60,6 +60,26 @@ static km_fork_state_t km_fork_state = {
...
+/*
+ * If this process has successfully fork()'ed, we don't allow snapshots to be taken.
+ * Snapshots must preserve unread data buffered in pipes and socketpairs and then write
+ * that data back into the other end of the pipe/socketpair.  If a fork
+ * has happened snapshot can't tell if a pipe or socketpair with a single end in this process
+ * has no other end, or the other end is in another process.
+ * If the other end is in another process, we can't snapshot buffered data.
+ * This counter is used by snapshot to disable snapshots after a fork has be done.
+ */
+unsigned int km_fork_count = 0;
```
gnode1, line 72: We might want to count children that exited, so even if there was fork but the child is gone we allow the snapshot.

paulpopelka:
Brought this PR up to the top of master.

km/km_fork.c
paulpopelka: What about children of children etc.

It might get to the point that we should just analyze processes that are descendants of the process being snapshotted.
If there are currently any descendants don't allow the snapshot.

gnode1:
Before the merge I'd like to make sure the condition when we discovered that doesn't happen any more. If I recall correctly it was one of node tests, but just to make sure I'd go over java and python as well

gnode1 (approved):
LGTM. You are going to rebase it and test it again, I assume. Let's also retest with the payloads that we used.

paulpopelka:
The previous push brings this branch up to the tip of master.

km/km_filesys.c
```diff
@@ -2077,41 +2077,76 @@ static inline int do_full_write(int fd, char* bufp, size_t bufl)
...
+      // The write side of this pipe is closed.  So we need to recreate the pipe
+      // to put the data back into the pipe buffer and then close the write end
+      // after writing the data.
+      // If any of this fails, we can not allow the payload to resume since we can't
+      // recreate is half open pipe or sockpair.
+      int fdpair[2];
+      if (file->how == KM_FILE_HOW_SOCKETPAIR0 || file->how == KM_FILE_HOW_SOCKETPAIR1) {
+         if (socketpair(file->sockinfo->domain, file->sockinfo->type, file->sockinfo->protocol, fdpair) <
+             0) {
+            km_err(1, "Unable to recreate half open socketpair, fd %d", fd);
```
gnode1, line 2108: should this be warn and return an error?

paulpopelka: No.  We have read data from a half open socketpair.  Now that we have read the data we need to recreate the socketpair, write the data back into it, and then close the end we've written the data into so that the payload will have the same half open socketpair it had before the snapshot was requested.  If any of these steps fail,  there is not much we can do but terminate the payload.

paulpopelka: Maybe you have another way to solve this?

gnode1 (approved):
LGTM, although I really haven't looked too deep into the logic

PR #1677, opened 2022-09-13
Change km_cli -s to work like it used to.

paulpopelka:
km_cli -s will now supply no snapshot file name to km. So km will use the value specified with its --snapshot command line argument.
Changed the bats continue_after_snapshot to work with this old way of doing things.

### DIFF
--- a/km_cli/km_client.c
+++ b/km_cli/km_client.c
@@ -17,7 +17,7 @@
 /*
  * Command line description:
  *
- * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l] [-t]
+ * km_cli [-c cmdname] [-p processid] [-d snapshotdir] [-s socket_name] [-l] [-t] [-r]
  *
  * There are 2 parts to this command, selection of processes to snapshot and then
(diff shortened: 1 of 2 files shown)


gnode1 (approved):

gnode1:
I think I'd prefer also reverse meaning of `-t` - terminate the process by default ...

johnmuth81 (approved):

km_cli/km_client.c
```diff
@@ -104,7 +104,7 @@ void usage(void)
 {
    fprintf(stderr,
            "Usage: %s [-l] [-c commandname] [-d snapshot_dirname] [-p processid] [-s "
-           "socket_name] [-t]\n",
+           "socket_name] [-t] [-x]\n",
```
gnode1, line 107: `-x`?

gnode1 (approved):
LGTM other then the typo in usage, and of course clang format

km_cli/km_client.c
paulpopelka: I was considering using the expire flag (x) but realized that didn't make sense.

PR #1678, opened 2022-09-14
Test payload snapshots

gnode1:
Adds node snapshot tests both on local machine and on CI.

Add java snapshot on local machine only. Java has no real tests on CI because it uses bats. Need to enable that in java directory.

Discovered bug in recovery of open sockets. We populate our filesys structure BUT we don't open file on the host. Next open file on the host will conflict with the our data. The test for this is Java snapshot test which I only ran locally (see above).

I'd really like to push this ASAP and figure out CI in Java after that.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -238,7 +238,7 @@ int km_add_guest_fd_internal(
                                    0,
                                    __ATOMIC_SEQ_CST,
                                    __ATOMIC_SEQ_CST) == 0) {
-      km_errx(0, "file slot %d is taken unexpectedly", host_fd);
+      km_errx(0, "file slot %d for %s is taken unexpectedly", host_fd, name);
    }
    km_file_t* file = &km_fs()->guest_files[host_fd];
    file->ops = ops;
@@ -2200,12 +2200,12 @@ static inline size_t fs_core_write_socket(char* buf, size_t length, km_file_t* f
    }
 
    km_infox(KM_TRACE_SNAPSHOT,
-            "fd:%d ISOCK: how=%d %s other=0x%x ofd %d, datalength %ld",
+            "fd:%d ISOCK: how=%d %s other=0x%x state=%d, datalength %ld",
             fnote->fd,
             fnote->how,
             file->name,
             fnote->other,
-            file->ofd,
+            fnote->state,
             fnote->datalength);
    return cur - buf;
 }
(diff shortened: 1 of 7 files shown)


sv641 (approved):

km/km_filesys.c
```diff
@@ -2660,13 +2650,13 @@ static int km_fs_recover_open_file(char* ptr, size_t length)
 
    int fd = open(name, nt_file->flags, 0);
```
paulpopelka, line 2651: codeql has decided we read a filename from user input here and then opening the file.

It seems like codeql has not gone deep enough to discover what km has opened and is reading the filename from.

We are reading from a snapshot file.  I guess codeql thinks that users manually generate snapshot files?

I guess we need to investigate what codeql is having fits about.

km/km_filesys.c
```diff
@@ -2537,7 +2537,13 @@ static inline int km_fs_recover_socket(km_nt_socket_t* nt_sock, struct sockaddr*
       return -1;
    }
    km_fs_recover_fd(nt_sock->fd, host_fd, 0, km_get_nonfile_name(host_fd), -1, nt_sock->how);
-
+   if (nt_sock->state == KM_NT_SKSTATE_ERROR) {
+      file->error = -ECONNRESET;
+      if (file->name != NULL) {
+         free(file->name);
+      }
+      file->name = strdup("socket error");
```
paulpopelka, line 2545: Check for allocation failure.

km/km_filesys.c
```diff
@@ -2537,7 +2537,13 @@ static inline int km_fs_recover_socket(km_nt_socket_t* nt_sock, struct sockaddr*
       return -1;
    }
    km_fs_recover_fd(nt_sock->fd, host_fd, 0, km_get_nonfile_name(host_fd), -1, nt_sock->how);
-
+   if (nt_sock->state == KM_NT_SKSTATE_ERROR) {
+      file->error = -ECONNRESET;
+      if (file->name != NULL) {
```
paulpopelka, line 2542: free() tolerates being passed a null pointer.  The man page even says so.

PS: this usually causes endless hours of developers having religious discussions over checking for null before calling free() or letting free() handle it.

km/km_filesys.c
gnode1: We can squelch this, tell it it's false positive

km/km_filesys.c
gnode1: We have a number of places when we do it this way, and the other way. Should we go over the whole things and make it uniform?

PR #1681, opened 2022-09-19
Change km_dump_core() to return an error on failure.

paulpopelka:
To allow a payload to continue running after a failed snapshot request we need to change km_dump_core to return an error instead of aborting/exiting km.  All of the functions called by km_dump_core were changed to return an error on failure.
Changed the callers of km_dump_core() to handle error returns.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -2435,19 +2435,18 @@ int km_fs_core_notes_write(char* buf, size_t length, size_t* sizep)
 /*
  * == Snapshot recovery
  */
-static inline int km_fs_recover_fd(int guestfd, int hostfd, int flags, char* name, int ofd, int how)
+static inline void km_fs_recover_fd(int guestfd, int hostfd, int flags, char* name, int ofd, int how)
 {
    km_file_t* file = &km_fs()->guest_files[guestfd];
 
    if (guestfd < 0) {
-      // Nothing to do for pipes and socketpairs where other end is closed.
       close(hostfd);
-      return 0;
+      return;
    }
    if (guestfd != hostfd) {
       if (guestfd != dup2(hostfd, guestfd)) {
          km_warn("can not dup2 %s to %d", name, guestfd);
-         return errno;
+         return;
       }
       close(hostfd);
    }
@@ -2466,7 +2465,6 @@ static inline int km_fs_recover_fd(int guestfd, int hostfd, int flags, char* nam
             name,
             ofd,
             how);
-   return 0;
 }
 
 static inline int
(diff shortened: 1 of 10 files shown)


km/km_filesys.c
```diff
@@ -2007,16 +2007,16 @@ size_t km_fs_core_notes_length(void)
             }
          } else if (file->sockinfo == NULL) {
             queuedbytes = 0;
-            if (file->how == KM_FILE_HOW_PIPE_1) {
+            if (file->how == KM_FILE_HOW_PIPE_0) {
```
gnode1, line 2010: Is this part of this PR?

km/km_filesys.c
```diff
@@ -2441,18 +2485,51 @@ km_fs_recover_fdpair(int guestfd[2], int hostfd[2], km_file_how_t how[2], int fl
...
    if (guestfd[0] == hostfd[1]) {
-      km_fs_recover_fd(guestfd[1], hostfd[1], flags[1], km_get_nonfile_name(hostfd[1]), guestfd[0], how[1]);
-      km_fs_recover_fd(guestfd[0], hostfd[0], flags[0], km_get_nonfile_name(hostfd[0]), guestfd[1], how[0]);
+      rc = km_fs_recover_fd(guestfd[1],
+                            hostfd[1],
+                            flags[1],
+                            km_get_nonfile_name(hostfd[1]),
+                            guestfd[0],
+                            how[1]);
+      if (rc != 0) {
```
gnode1, line 2502: So recover is changed the same way? Maybe not really needed, cannot really restart recover and hope to succeed.

gnode1 (commented):
Would be interesting to remove sleep from CI runs after node snapshot and see if that retry does help

km/km_filesys.c
paulpopelka: I got carried away.
I need to revert these recover changes.

paulpopelka:
Continuing from the previous comment, I guess we could scribble on the eventing fd to arrange for another event to be generated.
Some more thought is required to decide if this idea is feasible.

gnode1:
> > Would be interesting to remove sleep from CI runs after node snapshot and see if that retry does help
> 
> I think the reason the snapshot succeeded after the previous one failed is epoll_wait() is destructive. It harvests a pending event if there is one. We use this to decide if the snapshot can be taken. And then the event is gone so the payload won't be able to receive it. I need to find out if the eventfd info in /proc/self/fdinfo/XXX can be used to decide if we can snapshot the eventfd.

I think there is poll/select functionality to check for the state without destructing it

gnode1 (approved):

PR #1683, opened 2022-09-22
Change km to look for new env var KM_MGTDIR to config snapshotting.

paulpopelka:
The new environment variable KM_MGTDIR tells km where it should create its management pipe and where it should place its snapshots when asked to create one.
Management pipes are now created to be unique so that children of the main km process can create pipes whose names will not collide with the names of other pipes.  The pipename is created as follows:
  sprintf(pipename, "%s/kmpipe.%s.%d", $KM_MGTDIR,
          basename(km_payload_name), getpid());
Snapshots are place in KM_MGTDIR and are named in a similar fashion:
  sprintf(snapname, "%s/kmsanp.%s.%d", $KM_MGTDIR,
          basename(km_payload_name), getpid());

If the environment variable KM_MGTDIR exists and is non-null km will ignore the --mgtpipe, --snapshot command line flags and the KM_MGTPIPE environment variable.


paulpopelka:
I think the basic_snapshot failure:
```
# reply too small, got 0 bytes, expected 4 bytes
# Snapshot via management pipe /tmp/mgtdir.70297/kmpipe.hello_html_test.fedora.dyn.70895 failed, Invalid argument
is because we begin km exit processing before we even send the snapshot complete reply back to km_cli.
```
That area of the code needs to be rearranged.

km/km_management.c
```diff
@@ -96,14 +97,37 @@ static void* mgt_main(void* arg)
...
+         return;
+      }
    }
    if (strlen(path) + 1 > sizeof(addr.sun_path)) {
-      km_errx(2, "mgmt path too long");
+      km_errx(2, "mgmt path <%s> too long, max length %d", path, sizeof(addr.sun_path));
    }
    strncpy(addr.sun_path, path, sizeof(addr.sun_path) - 1);
 
+   km_warnx("snapshot pipe name is %s", path);
```
gnode1, line 129: info? I understand this is more for us the the customer, so I thought it makes sense to make it KM_VERBOSE controlled

km/km_management.c
```diff
@@ -96,14 +97,37 @@ static void* mgt_main(void* arg)
...
+      }
+      struct stat sb;
+      int rc = stat(km_mgtdir, &sb);
+      if (rc != 0) {
+         km_err(3, "KM_MGTDIR directory <%s> is not accessible", km_mgtdir);
+      }
+      snprintf(pipename, sizeof(pipename), "%s/kmpipe.%s.%d", km_mgtdir, basename(km_payload_name), getpid());
+      path = pipename;
+   } else {
+      if (path == NULL) {
```
gnode1, line 120: I'm a bit confused about the logic here. Are we blindly overwriting path in the clause above?

paulpopelka:
Forgot to address the memory scope problems which mgtpipe and the path variable being passed to km_mgt_init()  in main().
Will address next.

km/km_management.c
```diff
@@ -85,54 +94,86 @@ static void* mgt_main(void* arg)
...
 
-void km_mgt_init(char* path)
+void km_mgt_fini(void)
 {
-   if (path == NULL) {
+   unlink(addr.sun_path);
+   if (sock == -1) {
       return;
    }
+   kill_thread = 1;
```
gnode1, line 120: not needed is `sock == -1`?

km/km_management.c
```diff
@@ -85,54 +94,86 @@ static void* mgt_main(void* arg)
...
+   close(sock);
+   free(km_mgtdir);
```
gnode1, line 122: not needed is sock == -1?

km/km_management.c
paulpopelka: Im not sure what you are asking.
Do you think sock could be some other negative value than -1?
Do you think we are not setting sock to -1 when we should be?

gnode1 (approved):

johnmuth81 (approved):
lgtm with a nit

km/km_main.c
```diff
@@ -481,6 +495,12 @@ km_parse_args(int argc, char* argv[], int* argc_p, char** argv_p[], int* envc_p,
             km_set_snapshot_output_path(optarg);
             break;
          case 'm':
+            if (km_mgtdir != NULL) {
+               km_errx(1,
```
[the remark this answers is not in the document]
paulpopelka: > Why mutually exclusive instead of allowing command line to override environment variable?

Originally I had it so that the environment variable had priority over the command line flags and a warning was issued that the command line flag was being ignored.

Serge's thinking is: since we are probably running this in a container and since getting debug info out of a container is sometimes difficult it is better to just to fail immediately rather then letting the run continue with snapshot related values that are unexpected.

PR #1684, opened 2022-10-03
Ez leka/local dev

ez-leka:
No change in main-stream product. Strictly dev environment convenience. 
------------------------
Configure docker and podman to use development build of run and km (build/opt/kontain/bin)

```
make install-dev-runtime RUNTIME=<give it a distinct name>
```
To remove this configuration:
```
make uninstall-dev-runtime RUNTIME=<name from prev step>
```

Now docker and podman will try to use krun and km when the "--runtime your-runtime-name" argument is supplied with "docker run" or "podman run".
Without --runtime, runc will be the default container runtime.

To test payloads with development runtime
```
make -C payloads validate-runenv-image RUNTIME=your-runtime-name
```

### DIFF
--- a/container-runtime/docker_config.sh
+++ b/container-runtime/docker_config.sh
@@ -23,6 +23,26 @@ set -e ; [ "$TRACE" ] && set -x
 # This script must be run as root.
 [ `id -u` != "0" ] && echo "Must run as root" && exit 1
 
+# read arguments id any
+UNINSTALL=
+RUNTIME_NAME="krun"
+KRUN_PATH="/opt/kontain/bin/krun" 
+
+for arg in "$@"
+do
+   case "$arg" in
+        -u)
+            UNINSTALL=yes
+        ;;
+        --runtime-name=*)
+            RUNTIME_NAME="${1#*=}"
+        ;;
+        --runtime-path=*)
+            KRUN_PATH="${1#*=}"
+        ;;
+    esac
+    shift
+done
 # Can we assume docker is installed?
 #apt-get update
 #apt-get install -y -q docker.io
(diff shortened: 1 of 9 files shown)


container-runtime/docker_config.sh
```diff
@@ -23,6 +23,26 @@ set -e ; [ "$TRACE" ] && set -x
...
+do
+   case "$arg" in
+        -u)
+            UNINSTALL=yes
+        ;;
+        --runtime-name=*)
+            RUNTIME_NAME="${1#*=}"
+        ;;
+        --runtime-path=*)
+            KRUN_PATH="${1#*=}"
```
paulpopelka, line 41: Could you verify that KRUN_PATH (the file itself) exists.
It is a pain trying to discover this when docker and its helpers fail because of this.

container-runtime/podman_config.sh
```diff
@@ -32,6 +32,24 @@ set -e ; [ "$TRACE" ] && set -x
...
+   case "$arg" in
+        -u)
+            UNINSTALL=yes
+        ;;
+        --runtime-name=*)
+            RUNTIME_NAME="${1#*=}"
+        ;;
+        --runtime-path=*)
+            KRUN_PATH="${1#*=}"
+            KM_PATH=$(echo "${KRUN_PATH}" | sed 's/krun/km/')
```
paulpopelka, line 48: Is this really used somewhere?

smijar (approved):

paulpopelka:
It looks like the azure and aws runners need the jq program added to their images.

jq-1.6-10.fc34.x86_64

or whatever is appropriate for those versions of linux.

One other set of tests is failing for reasons I don't understand.

paulpopelka:
I think the place to add jq to the list of packages is in:
```
./cloud/scripts/L0-image-provision.sh
```
This statement:
```
apt-get install --yes -q git make makeself gcc linux-headers-$(uname -r) \
                         docker-ce docker-ce-cli containerd.io azure-cli \
                         vagrant packer virtualbox
```

container-runtime/docker_config.sh
ez-leka: done

PR #1686, opened 2022-10-06
Change gdb_lots_of_threads.c to wait for all threads to be running

paulpopelka:
I have found that threads created by pthread_create() are not running immediately and can't be pthread_join()'ed until they are running. This fix is being put in because the gdb_attch bats test failed because gdb info threads reported all threads were on vcpu-0. I don't understand how that would happen but I think those new threads were not ready to be debugged because pthread_create() for the vcpu threads in km were not done with their work.
Since this test does do pthread_join() we should wait for the threads to be running before doing the join so this change doesn't hurt.

### DIFF
--- a/tests/gdb_lots_of_threads_test.c
+++ b/tests/gdb_lots_of_threads_test.c
@@ -45,6 +45,7 @@ time_t starttime;
 int stop_after_seconds;
 int wait_for_gdb = 1;
 int running_count = 0;
+int all_running = 0;
 
 pthread_mutex_t rand_serialize = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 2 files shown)


tests/gdb_lots_of_threads_test.c
```diff
@@ -166,6 +168,13 @@ int main(int argc, char* argv[])
       }
    }
 
+   // Wait for all the threads to be running.
+   // Yes, sometimes the threads have been created but are not yet joinable.
+   while (running_count < i) {
```
gnode1, line 173: You want to make this into atomic fetch? Maybe it doesn't matter on x86

gnode1 (approved):

tests/gdb_lots_of_threads_test.c
paulpopelka: ok, I change it.

paulpopelka:
> The tests fail. I guess gdb also needs to wait fot the threads to start before looking for things in them
I thought that is what the change did.

gnode1 (approved):

PR #1691, opened 2022-10-18
Add debug for the bats continue_after_snapshot test

paulpopelka:
The continue_after_snapshot bats test occasionally fails with this message:
```
07:12:47.612374 km_add_guest_fd_inte 241  vcpu-0  file slot 1 for (null) is taken unexpectedly
```
We don't know what is wrong here.  To maybe get some clues I've changed km's options to generate debug tracing into a file for this test.
I've also changed the km code that generates this message to abort instead of just exiting with a message.

Hopefully in a week or 2 we will run into the problem again and there will be some trace data to look at.

### DIFF
--- a/km/km.h
+++ b/km/km.h
@@ -497,6 +497,7 @@ extern const_string_t km_cmd_short_options;
 
 void km_trace_fini(void);
 void km_trace_setup(int argc, char* argv[]);
+void km_pathetic_stacktrace(void);
 
 extern int km_collect_hc_stats;
 
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -233,12 +233,8 @@ int km_add_guest_fd_internal(
    int available = 0;
    int taken = 1;
    km_file_t* file = &km_fs()->guest_files[host_fd];
-   if (__atomic_compare_exchange_n(&file->inuse,
-                                   &available,
-                                   taken,
-                                   0,
-                                   __ATOMIC_SEQ_CST,
-                                   __ATOMIC_SEQ_CST) == 0) {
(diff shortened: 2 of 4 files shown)


gnode1 (approved):
I thought you were planning to use `backtrace()` and friend to put the stack trace in the CI log. The way its done now the core will disappear with the github runner

paulpopelka:
> I thought you were planning to use `backtrace()` and friend to put the stack trace in the CI log. The way its done now the core will disappear with the github runner

I played around with backtrace() et al and discovered it really didn't do much that I thought was useful.
It looks like you need to use the addr2line program and the addresses from backtrace() to give file and line number information.
I guess they decided adding all of that capability was not worth the memory consumption trade-off.  I don't know.

gnode1:
Thing is that we'll not see the core file if it fails on github runner,
where it tends to fail ...

paulpopelka:
> Thing is that we'll not see the core file if it fails on github runner, where it tends to fail ...

But we will get a km trace, if may bats hackery works.

I can work on making backtrace plus whatever is in the addr2line program a function to call from km then we can get a usable traceback.

paulpopelka:
Note that for the backtrace produced by backtrace() et al to have symbols we need to link km with the -rdynamic flag which causes km's external symbols that are only referenced internally (by km) to be placed in km's symbol table.
Other stack unwinders are willing to use debug symbols to produce the stack trace.  Then we would need to link km with a new library.  I'm sure we will need to have a group discussion about that.

This link discusses -rdynamic with an interesting article siting reasons to not use it.  Not strong reasons but some reasons.
https://stackoverflow.com/questions/36692315/what-exactly-does-rdynamic-do-and-when-exactly-is-it-needed

PR #1693, opened 2022-10-24
lightweight snap restore - wait for connection before restoring snapshot

gnode1:
In restoring snapshot, if `SNAP_LISTEN_PORT` environment variable is set, before doing anything listen on port defined by the `SNAP_LISTEN_PORT` using select. Select reports readable event whet connection arrives, but keeps it queued. Initialize VM and payload after that, taking care of restoring the listening socket to the right state. 

Since the arriving connection is still queued no need to do anything specific about that, the payload will accept it regularly.

Tested manually.

Known issues:

- The listening socket is ipv4, so the snapshot has to be taken on ipv4, otherwise it doesn't work. In my testing disabled ipv6 on the machine. Going forward we'll probably need a parameter to tell km which domain to use.
- Same applies to the address - right now it is INADDR_ANY. If payload uses specific IP to listen on it might cause an issue.
- We only support one listening socket. If payload listens on several socket using select/poll we'll need to follow the suite and establish several sockets to listen.


sv641 (approved):
code looks good.
light_snap_listen could move to a separate file.

km/km_filesys.c
```diff
@@ -2624,10 +2677,46 @@ static inline int km_fs_recover_socket(km_nt_socket_t* nt_sock, struct sockaddr*
...
    km_assert(km_is_file_used(file) == 0);
 
-   int host_fd = socket(nt_sock->domain, nt_sock->type, nt_sock->protocol);
-   if (host_fd < 0) {
-      km_warn("socket recover");
-      return -1;
+   int host_fd;
+   // check if this is lightweight snap listening socket, retore it accordingly
+   if (snap_listen_sockinfo != NULL && nt_sock->domain == snap_listen_sockinfo->domain &&
+       memcmp(nt_sock + 1, snap_listen_sockinfo->addr, snap_listen_sockinfo->addrlen)) {
```
paulpopelka, line 2683: Should this be memcmp(....) == 0?
Don't you want to check for a matching bind address?
I guess since the domains are the same we don't need to compare  nt_sock->addrlen with snap_listen_sockinfo->addrlen for equality.

km/km_filesys.c
```diff
@@ -2624,10 +2677,46 @@ static inline int km_fs_recover_socket(km_nt_socket_t* nt_sock, struct sockaddr*
    km_file_t* file = &km_fs()->guest_files[nt_sock->fd];
    km_assert(km_is_file_used(file) == 0);
 
-   int host_fd = socket(nt_sock->domain, nt_sock->type, nt_sock->protocol);
-   if (host_fd < 0) {
-      km_warn("socket recover");
-      return -1;
+   int host_fd;
+   // check if this is lightweight snap listening socket, retore it accordingly
```
paulpopelka, line 2681: restore not retore

km/km_filesys.c
paulpopelka: I guess you can listen again on the same port + addr and the kernel recognizes 2 listeners on different fd's for the same address.

paulpopelka: SO_REUSEADDR helped too I think.

km/km_filesys.c
```diff
@@ -75,6 +77,58 @@ static int km_fs_g2h_readlink(const char* name, char* buf, size_t bufsz);
 km_fd_dup_data_t dup_data;
 pthread_mutex_t dup_data_mtx;
 
+/*
+ * Lightweight snap start - wait for connection on the snap_listen_sock before restoring the snapshot.
+ */
+static const_string_t KM_SNAP_LISTEN_PORT = "SNAP_LISTEN_PORT";
```
johnmuth81, line 83: IMO we need to centralize these environment variables in a place where we can provide some in-line documentation.

km/km_main.c
```diff
@@ -645,6 +637,16 @@ int main(int argc, char* argv[])
       if (argc_p > 1) {
          km_errx(1, "cannot set payload arguments when resuming a snapshot");
       }
+      light_snap_listen();
```
johnmuth81, line 640: There should be a command line option way to trigger this too I think.

johnmuth81 (commented):
Cool technique.

km/km_filesys.c
gnode1: Agree, but I'd say it's a separate PR

km/km_filesys.c
```diff
@@ -75,6 +77,58 @@ static int km_fs_g2h_readlink(const char* name, char* buf, size_t bufsz);
 km_fd_dup_data_t dup_data;
 pthread_mutex_t dup_data_mtx;
 
+/*
```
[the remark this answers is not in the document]
gnode1: It interacts pretty closely with the `km_fs_recover_socket()`, so I thought it makes sense to put it here. It is part of snapshot recovery process, which is heavily present here

paulpopelka:
I can't believe I'm saying this.  There needs to be a basic bats test for this to ensure we don't break it with future changes.
Sorry.

gnode1:
You are right of course ...

km/km_filesys.c
```diff
@@ -75,6 +77,56 @@ static int km_fs_g2h_readlink(const char* name, char* buf, size_t bufsz);
...
+   char* port;
+   if ((port = getenv(KM_SNAP_LISTEN_PORT)) == NULL) {
+      return;
+   }
+   snap_listen_sock = socket(AF_INET, SOCK_STREAM, 0);
+
+   snap_listen_sockinfo = malloc(sizeof(km_fd_socket_t));
+   *snap_listen_sockinfo = (km_fd_socket_t){.domain = AF_INET, .type = SOCK_STREAM, .protocol = 0};
+
+   struct sockaddr_in sa_serv = {.sin_family = AF_INET, .sin_addr.s_addr = INADDR_ANY};
```
johnmuth81, line 100: See this document: https://tldp.org/HOWTO/Linux+IPv6-HOWTO/ch23s01.html.
In particular, see the example in section `1.4.1.  'Daytime' Server Code'`

Note how `getaddrinfo` is used to get a list of constructed socket addresses, both IPv4 and IPv6, to be used by `bind`. This is the way.

gnode1: I tried that example program - the end result boils down to the same thing. Granted your way is the better way but it ain't my problem ATM.

km/km_coredump.c
```diff
@@ -959,6 +962,51 @@ int km_dump_core(char* core_path,
       }
    }
 
+   if (dumptype == KM_DO_SNAP) {
+      int cfd;   // conf file fd
+      char cpath[MAXPATHLEN];
+      sprintf(cpath, "%s.conf", core_path);
+      if ((cfd = open(cpath, O_RDWR | O_CREAT | O_TRUNC, 0600)) < 0) {
+         km_warn("Cannot create %s '%s'", "snapshot config", cpath);
+         return errno;
```
paulpopelka, line 971: You should "goto out;" here instead.
That will free memory and close the open core file fd.
And, we should probably delete the snapshot if we are failing the snapshot (that is not there currently).

km/km_snapshot.c
```diff
@@ -397,13 +397,6 @@ static inline int km_ss_recover_vcpus(char* notebuf, size_t notesize)
 static inline int km_ss_recover_km_monitor(char* notebuf, size_t notesize)
 {
    km_nt_monitor_t* mon = (km_nt_monitor_t*)notebuf;
-   char* base = (char*)(mon + 1);
-   if (mon->label_length > 0) {
-      km_warnx("label: %s", base);
-   }
-   if (mon->description_length > 0) {
-      km_warnx("description: %s", base + mon->label_length);
-   }
```
paulpopelka, line 406: I assume this was removed because it is annoying.
Nothing is fixed here, right?

km/km_coredump.c
```diff
@@ -968,7 +968,8 @@ int km_dump_core(char* core_path,
       sprintf(cpath, "%s.conf", core_path);
       if ((cfd = open(cpath, O_RDWR | O_CREAT | O_TRUNC, 0600)) < 0) {
          km_warn("Cannot create %s '%s'", "snapshot config", cpath);
-         return errno;
+         (void)unlink(cpath);
```
paulpopelka, line 971: I meant for the unlink() to be done down at the out: label. Sorry I wasn't clear.
unlinking here seems pointless because the open failed.

paulpopelka:
Looks like the CI tests are failing because this:
```
assert [ -f ${MGTDIR}/kmsnap.hello_html_test$ext.[0-9]* ]
```
is expanding to this:
```
# expression : [ -f /tmp/mgtdir.40890/kmsnap.hello_html_test.fedora.dyn.41421 /tmp/mgtdir.40890/kmsnap.hello_html_test.fedora.dyn.41421.conf ]
```
and upsetting the test command.

paulpopelka:
Sorry, I didn't mean to close this.

km/km_snapshot.c
gnode1: These were forcing the `/tmp/km_log.1` with every snapshot run, destroying the open on demand logic you did a while back. And I see no value. Also Asked John, he confirmed it was debug cruft

km/km_filesys.c
```diff
@@ -1649,7 +1670,27 @@ uint64_t km_fs_accept(km_vcpu_t* vcpu, int sockfd, struct sockaddr* addr, sockle
...
+      snap_listen_sock = -1;
+   } else {
+      if ((hostfd = __syscall_4(SYS_accept4, host_sockfd, (uintptr_t)addr, (uintptr_t)addrlen, fl)) < 0) {
+         return hostfd;
+      }
+   }
+   if (fl != 0) {
+      int tmp = fcntl(hostfd, F_GETFL, NULL);
+      fcntl(hostfd, F_SETFL, tmp | (fl & (SOCK_NONBLOCK | SOCK_CLOEXEC)));
+   }
```
paulpopelka, line 1693: This should only be done if snap_listen_sock == sockfd.
For the normal case, the accept system call will have done it.

km/km_filesys.c
```diff
@@ -26,6 +26,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <netdb.h> /* getaddrinfo(3) et al.                       */
```
paulpopelka, line 29: Is this still needed?

km/km_filesys.c
```diff
@@ -137,17 +141,34 @@ void light_snap_listen(void)
       km_err(2, "listen failed");
    }
 
-   // wait for connection to arrive but don't accept it just yet. When payload starts it will accept it
-   fd_set fds;
-   FD_ZERO(&fds);
-   FD_SET(snap_listen_sock, &fds);
+   do {
+      snap_conn_sock =
```
paulpopelka, line 145: Error check needed here?

paulpopelka:
One other think that needs some attention.  It is tangential to the change so you can chose to ignore.

accept_hcall() does not handle the addr and addrlen address arguments being null.  It should.

km/km_hcalls.c
```diff
@@ -1971,7 +1971,7 @@ static km_hc_ret_t snapshot_putdata_hcall(void* vcpu, int hc, km_hc_args_t* arg)
 
 static km_hc_ret_t shrink_payload_hcall(void* vcpu, int hc, km_hc_args_t* arg)
 {
-   int km_shrink_footprint(km_vcpu_t* vcpu);
+   int km_shrink_footprint(km_vcpu_t * vcpu);
```
paulpopelka, line 1974: The one place I get our coding style correct and it is changed to something different?

paulpopelka: This prototype can be removed since it is now in km.h.

tests/km_core_tests.bats
```diff
@@ -1197,6 +1198,27 @@ fi
...
+   pid=$!
+   curl localhost:$snapshot_test_port
+   wait $pid
+   assert_success
+   assert grep -q "select() returned that input is available on listening fd" ${MGTDIR}/prelisten.log
+   assert grep -q "poll() returned that input is available on listening fd " ${MGTDIR}/prelisten.log
+   assert grep -q "epoll_wait() returned that input is available on listening fd" ${MGTDIR}/prelisten.log
+   assert grep -q "accept returned fd" ${MGTDIR}/prelisten.log
+   rm -fr ${MGTDIR}
+
```
paulpopelka, line 1221: You should remove all of this since it will break CI until I fix it.

paulpopelka: I've pushed a change that fixes this.

paulpopelka (approved):

docs/light_snapshot_start.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Making the snapshot
+
+Preparing the snapshot for this feature is the same process as for regular snapshot.
+In addition to `kmsnap` file km now creates another file called `kmsnap.conf`.
+This file records the info about the listening socket.
+It looks like this:
+
+```txt
+i6 8080 ::
```
smijar, line 22: what is the i6

docs/light_snapshot_start.md
```diff
@@ -0,0 +1,48 @@
...
+
+Once the payload is triggered it serves the triggering request, and keeps running after that.
+If any new request comes it will serve it too.
+
+In order to take full advantage an idle service should scale to zero.
+We achieve that by a timeout on the incoming requests.
+The timeout is specified by another environment variable setting the timeout in milliseconds:
+
+```bash
+docker run ... --env=SNAP_LISTEN_TIMEOUT=1000 --env=SNAP_LISTEN_PORT="i6 8080 ::" ...
```
smijar, line 45: ability to configure autoshrink is good.

smijar (approved):

docs/light_snapshot_start.md
paulpopelka: IP v6 address family.

paulpopelka (approved):

Issue #1661, opened 2022-08-19
snapshot recovery fails when one of eventfd's listeners is in accept state


sv641:
I took a snapshot of go server provided by knative examples. snapshot created fails to start with error

`[h1@kubernetes km]$ ./build/km/km ../snap/kmsnap 
18:47:10.204619 km_ss_recover_km_mon 401  km      label: /server 1 /tmp/kmmgmt.pipe-1
18:47:10.204755 km_ss_recover_km_mon 404  km      description: snapshot date Thu Aug 18 18:38:07 2022

18:47:10.208519 km_fs_recover_eventf 2966 km      monitored fd=5 does not exist
18:47:10.208529 km_fs_recover        2995 km      recover open files failed
[h1@kubernetes km]$ 
`

sv641:
Further debugging the snapshot recover with lots of help from @paul we found that when socket is in

`   if (nt_sock->how == KM_FILE_HOW_ACCEPT) {
      return km_fs_recover_socket_accepted(nt_sock);
   }
`
`static int km_fs_recover_socket_accepted(km_nt_socket_t* nt_sock)
{
   return 0;
}`

In memory filesys entry is not being made and this is causing eventfd creation to bail out.

sv641:
fd state of process being snapshoted
`lrwx------    1 root     root            64 Aug 18 18:35 0 -> /dev/null
l-wx------    1 root     root            64 Aug 18 18:35 1 -> pipe:[110166]
l-wx------    1 root     root            64 Aug 18 18:35 2 -> pipe:[110167]
lrwx------    1 root     root            64 Aug 18 18:35 3 -> socket:[109236]
lrwx------    1 root     root            64 Aug 18 18:35 5 -> socket:[109260]
lrwx------    1 root     root            64 Aug 18 18:35 6 -> anon_inode:[eventpoll]
lrwx------    1 root     root            64 Aug 18 18:38 729 -> socket:[114797]
l-wx------    1 root     root            64 Aug 18 23:05 731 -> /tmp/km_1.log
lrwx------    1 root     root            64 Aug 18 23:05 732 -> anon_inode:[eventfd]
lrwx------    1 root     root            64 Aug 18 23:05 733 -> anon_inode:[eventfd]
lrwx------    1 root     root            64 Aug 18 23:05 734 -> /dev/kvm
lrwx------    1 root     root            64 Aug 18 23:05 735 -> anon_inode:kvm-vm
[...]

PR #1694, opened 2022-10-27
Fix snapshot recovery of established network connections, and more

paulpopelka:
This change started out to fix the problem of connections that are establish when a snapshot is taken should be recovered as unconnected and return an error when a system call using the now disconnection connection. That is fixed but more stuff ended up being piled in here. The other stuff fixed is:

- clear the error field of the km_file_t when a file is opened. It should probably be cleared after an error is returned but this fix is already pretty big.

- Add tests to tests/snapshot_test.c to verify that socket bound addresses are being recovered.

- Reduce the number of values used to describe how an fd came into existence and the state of a socket.  Formerly there were values stored in the km_file_t an the km_fd_socket_t and in the snapshot elf notes.  Now we only use the km_file_t and km_fd_socket_t values in km_file_t and the elf notes.

- a socket with a local address bound is not really in a different state.  We were representing such sockets as being in a different state.  Being bound is just a property of a socket.  If you read the ip(7) man page you will find that binding an address is not required for a listen to be performed and you can bind a local address to a socket that will never be performing a connect.  The code in snapshot is changed to treat the bound local address as a socket property now.

- the shutdown() system call changes the "connected" state of a socket. km was not handling this at all.  It does now.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -249,6 +249,7 @@ int km_add_guest_fd_internal(
       file->name = strdup(name);
    }
    file->flags = flags;
+   file->error = 0;
    return host_fd;
 }
 
@@ -268,7 +269,10 @@ static inline int km_add_socket_fd(
 {
    int ret = km_add_guest_fd_internal(vcpu, hostfd, name, flags, how, NULL);
    if (ret >= 0) {
-      km_fd_socket_t sockval = {.domain = domain, .type = type, .protocol = protocol};
+      km_fd_socket_t sockval = {.state = KM_SOCK_STATE_UNCONNECTED,
+                                .domain = domain,
+                                .type = type,
+                                .protocol = protocol};
       km_fd_socket_t* sockinfo = calloc(1, sizeof(km_fd_socket_t));
       *sockinfo = sockval;
       km_fs()->guest_files[ret].sockinfo = sockinfo;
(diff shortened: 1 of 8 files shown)


paulpopelka:
Fixes PR #1661

paulpopelka:
A note to myself for the flleeting exit thread test failure:
```
not ok 365 threads_exit_grp(glibc_static): force exit when threads are in flight (exit_grp_test.fedora)
# (from function `assert' in file bats-assert/src/assert.bash, line 46,
#  in test file km-bats-7.glibc_static, line 641)
#   `assert [ $status -eq 17 -o $status -eq 11  ]' failed with status 134
# start: 2022-10-27 16:59:25.185591119+00:00 threads_exit_grp(glibc_static): force exit when threads are in flight (exit_grp_test.fedora)
# --- Test script output:
# 
# -- assertion failed --
# expression : [ 134 -eq 17 -o 134 -eq 11 ]
# --
# 
#   end: 2022-10-27 16:59:28.621158857+00:00 threads_exit_grp(glibc_static): force exit when threads are in flight (exit_grp_test.fedora)
# 
# --- Command line:
# km_with_timeout exit_grp_test.fedora
# --- Command output:
# .16:59:28.184711 km_dump_core         875  vcpu-9  Writing coredump to './kmcore'
# 16:59:28.583468 km_deliver_signal    624  vcpu-9  guest: Terminated by signal: Aborted (core dumped)
# 
# Killed by signal. Timeout returns 134: /home/appuser/km/build/opt/kontain/bin/km --virt-device=/dev/kvm --km-log-to=stderr   exit_grp_test.fedora
# ---
```

km/km_filesys.c
```diff
@@ -1605,7 +1639,7 @@ uint64_t km_fs_connect(km_vcpu_t* vcpu, int sockfd, struct sockaddr* addr, sockl
    ret = __syscall_3(SYS_connect, host_sockfd, (uintptr_t)addr, (uintptr_t)addrlen);
    if (ret >= 0) {
       km_fd_socket_t* sock = km_fs()->guest_files[sockfd].sockinfo;
-      sock->state = KM_SOCK_STATE_CONNECT;
+      sock->state = KM_SOCK_STATE_CONNECTED;
```
gnode1, line 1642: Same in accept4?

km/km_filesys.c
```diff
@@ -507,7 +511,34 @@ int km_fs_shutdown(km_vcpu_t* vcpu, int sockfd, int how)
    if (ret != 0) {
       return ret;
    }
+   km_fd_socket_t* sock = km_fs()->guest_files[host_fd].sockinfo;
    ret = __syscall_2(SYS_shutdown, host_fd, how);
+   if (ret == 0 && sock->state != KM_SOCK_STATE_UNCONNECTED) {
```
gnode1, line 516: I wonder if we should do this for ENOTCONN?

km/km_filesys.c
```diff
@@ -1171,10 +1202,14 @@ uint64_t km_fs_fstat(km_vcpu_t* vcpu, int fd, struct stat* statbuf)
    if ((host_fd = km_fs_g2h_fd(fd, NULL)) < 0) {
       return -EBADF;
    }
+#if 0
```
gnode1, line 1205: debug cruft?

km/km_filesys.c
```diff
@@ -1587,7 +1621,7 @@ uint64_t km_fs_accept(km_vcpu_t* vcpu, int sockfd, struct sockaddr* addr, sockle
       return -ENOMEM;
    }
    km_fd_socket_t* sock = km_fs()->guest_files[guestfd].sockinfo;
-   sock->state = KM_SOCK_STATE_ACCEPT;
+   sock->state = KM_SOCK_STATE_CONNECTED;
```
gnode1, line 1624: Same in accept4?

gnode1 (changes requested):
Generally looks good, there is one question about accept4, and a couple of nits

paulpopelka:
Please ignore all that stuff in the review I apparently started.

paulpopelka:
This change was absorbed into another change which has been merged.
This will be closed as not needed.

PR #1699, opened 2022-11-15
don't shrink if there is connected socket

gnode1:
keep track of the latest connected socket and don't trigger payload shrink if it is still open


sv641 (approved):

paulpopelka (approved):
This is like pulling a string.  We find more stuff as we keep pulling.

gnode1:
Significantly redone. We now maintain the count of active accepted sockets, ++ for each accept, -- for close. Use the count to determine if we can shrink the payload. If the count is zero then check for timeout since the last accept.

Also, check it again after the VCPUs are stopped, to make sure no new accepts snick in between

sv641 (approved):

PR #1701, opened 2022-11-23
Give better error when the SNAP_LISTEN_PORT env var can't be decoded

paulpopelka:
light_snap_listen() decodes what is stored in the SNAP_LISTEN_PORT env var.  If it can't decide the variable, it just drops thru to the listen call with an invalid fd which causes listen() to fail. This change adds code to print a message showing the contents of SNAP_LISTEN_PORT before exiting km.


paulpopelka:
On of the CI tests fails occasionally with listen on fd -1.
I think the problem is SNAP_LISTEN_PORT has a bad value.
Maybe we can figure out what is wrong if we can see the variable contents.

gnode1 (approved):

PR #1705, opened 2022-11-29
build issues with gcc 11.3.0

sv641:


km/km_coredump.c
```diff
@@ -1001,14 +1001,18 @@ int km_dump_core(char* core_path,
                         break;
                   }
                   strcat(buf, "\n");
-                  write(cfd, buf, strlen(buf));
+                  if (write(cfd, buf, strlen(buf)) != strlen(buf)) {
```
gnode1, line 1004: If you really want to go all the way here - check for errno, and if it's zero, loop untill all is written. Note `int do_full_write(int fd, char* bufp, size_t bufl)` function in km_filesys.c

gnode1:
I'm wondering why I have no issues like these with gcc version 12.2.1

PR #1711, opened 2022-12-19
Improve snapshot messages and uninterpreted opcode message

paulpopelka:
Todays CI failure shows a km trace that has both a snapshot generation and snapshot resume all run together.  To help us realize what is happening the next time we see this there are km traces now for begin and end of snapshot generation and resumption.

Plus the message for an uninterpreted opcode has been changed to include both the PC for the instruction and the first 6 bytes of the instruction.

### DIFF
--- a/km/km_decode.c
+++ b/km/km_decode.c
@@ -820,7 +820,16 @@ static void decode_opcode(km_vcpu_t* vcpu, x86_instruction_t* ins)
       return;
    }
 
-   km_warnx("KM intruction decode: Uninterpreted Opcode=0x%x", opcode);
+   uint64_t pc = vcpu->regs.rip;
+   km_warnx("KM intruction decode: PC 0x%x, Uninterpreted Instruction=%02x %02x %02x %02x %02x "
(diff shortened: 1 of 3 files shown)


km/km_decode.c
```diff
@@ -820,7 +820,16 @@ static void decode_opcode(km_vcpu_t* vcpu, x86_instruction_t* ins)
       return;
    }
 
-   km_warnx("KM intruction decode: Uninterpreted Opcode=0x%x", opcode);
+   uint64_t pc = vcpu->regs.rip;
+   km_warnx("KM intruction decode: PC 0x%x, Uninterpreted Instruction=%02x %02x %02x %02x %02x "
```
sv641, line 824: not enough bytes for some instructions. But we will get the prefix which is what we need.

sv641 (approved):

km/km_decode.c
paulpopelka: There should be enough for prefix, opcode, modrm byte, and the sib byte.

PR #1712, opened 2022-12-20
light snap start with payload shrink on the last accepted socket close

gnode1:
Instead of using timeout as indicator of payload be done. use only accepted socket count, shrink the payload when that drops to zero.

The old behavior is preserved - if timeout is positive it is treated as timeout in milliseconds. But if it is negative the counter only approach is used.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -82,7 +82,9 @@ pthread_mutex_t dup_data_mtx;
 /*
  * Lightweight snap start - wait for connection on the snap_listen_sock before restoring the snapshot.
  */
-u_int64_t light_snap_accept_timeout;
+int64_t light_snap_accept_timeout;   // != 0 means light weight accept is enabled
+                                     // < 0 means shrink only by accept count
+                                     // > 0 means shrink by accept count and timeout
 static int snap_listen_sock = -1;
 static km_fd_socket_t* snap_listen_sockinfo;
 static int snap_conn_sock = -1;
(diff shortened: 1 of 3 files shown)


sv641 (approved):
How are dup's handled?

johnmuth81 (approved):

km/km_filesys.c
```diff
@@ -93,15 +95,29 @@ static u_int64_t accept_time;   // time stamp of the latest accept HC in millise
...
    if (light_snap_accept_timeout != 0) {
       __atomic_add_fetch(&accept_cnt, 1, __ATOMIC_SEQ_CST);
-      struct timespec tp;
-      if (clock_gettime(CLOCK_MONOTONIC, &tp) != 0) {
-         km_err(2, "can't update accept time");
+      if (light_snap_accept_timeout > 0) {
+         struct timespec tp;
+         if (clock_gettime(CLOCK_MONOTONIC, &tp) != 0) {
+            km_err(2, "can't update accept time");
+            accept_time = tp.tv_sec * 1000 + tp.tv_nsec / 1000000;
```
paulpopelka, line 106: This statement should be outside the "if" block.
km_err() exists, so it would never be reached anyway.

gnode1: Ha! That maybe why I see no difference with or without snapshots. Good catch.

gnode1:
> How are dup's handled?

There is no differece in dup han

gnode1:
> How are dup's handled?

It's a good question. This change is independent of dups. Even with timeout we also depend on the count. This change simply allows no timeout, or rather very short timeout without spinning.

The count logic ignores dups which is not right. I guess we should ++ on dup, then it should be right.

PR #1713, opened 2022-12-22
Add support for vex instruction prefix to km_decode.c

paulpopelka:
Addeded support for the VEX 2 and 3 byte instruction prefixes. Also added a smoke test for these prefixes to the bats decode_test. Since the XOP prefix is similar to the VEX 3 byte prefix, it is plumbed in but its tables of opcodes that reference memory has not been filled in.

### DIFF
--- a/km/km_decode.c
+++ b/km/km_decode.c
@@ -854,7 +854,7 @@ static void decode_opcode(km_vcpu_t* vcpu, x86_instruction_t* ins)
  */
 #define VEX_2BYTE_PREFIX 0xc5
 #define VEX_3BYTE_PREFIX 0xc4
-#define XOP_PREFIX       0x8f
+#define XOP_PREFIX 0x8f
 #if 0
 // For computing the fault address, we ignore these prefixes.
 // This is just here to remind us that the vex pp field can be ignored.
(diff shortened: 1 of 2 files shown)


gnode1 (approved):
LGTM

Maybe change the other tables to use `static const u_int8_t` instead of `unsigned char`?

km/km_decode.c
```diff
@@ -834,6 +860,488 @@ static void decode_opcode(km_vcpu_t* vcpu, x86_instruction_t* ins)
...
+// In the amd manual, tables A-25 through A-31 specify the xop opcode maps
+// and the xop opcode groups
+const unsigned char* xop_groups[32];
+const ushort* xop_maps[32];   // 8-10 are valid indexes
+
+static unsigned short find_opcode_in_map(const ushort* map, unsigned char opcode)
+{
+   const ushort* p;
+
+   for (p = map; *p != OPCODE_MAP_END; p++) {
```
sv641, line 1216: are you sure you want to search a table on each fault?

sv641 (approved):

km/km_decode.c
paulpopelka: When exception processing becomes a normal programming technique then we will need to change.
Given how people seem to think it wouldn't surprise me if that is what the industry decided to do.

paulpopelka:
> LGTM
> 
> Maybe change the other tables to use `static const u_int8_t` instead of `unsigned char`?

Are you referring to all of the tables in the km_decode.c source file, or just my changes?

gnode1:
> > LGTM
> > Maybe change the other tables to use `static const u_int8_t` instead of `unsigned char`?
> 
> Are you referring to all of the tables in the km_decode.c source file, or just my changes?

I was talking about other tables. Yours look good now

PR #1714, opened 2022-12-28
detailed bin artifact report

gnode1:
Print more details about kontan_bin.tar.gz artifact to diagnose failures


paulpopelka (commented):
Is there a way to put the version number of upload-artifact in a macro so that we don't need to make this change all over when a bug in the version we use surfaces?

Is it necessary to include a version?  Is the latest version of upload-artifact broken?

Just questions from the ignorant.

gnode1:
> Is there a way to put the version number of upload-artifact in a macro so that we don't need to make this change all over when a bug in the version we use surfaces?
> 
> Is it necessary to include a version? Is the latest version of upload-artifact broken?
> 
> Just questions from the ignorant.

All I know is that in all of the examples and tutorials for github actions they do specify version number and use implicit numbers for that.

If it were something we control we could've use some macro expansion facility (like m4) to generate the .yaml file from a template. But this is all processed by github strictly based on the checked in file content. 

I'll try the non-versioned actions, hopping that would default to the last version. It of course has another weakness - what it parameters change between the versions?

ez-leka (approved):

PR #1720, opened 2023-01-16
Add error checking to code that writes snapshot conf file

paulpopelka:
And add test in bats that verifies the kmsnap.*.conf file exists and is non-zero length.

This change is to help narrow down the problem we see once in a while when the bats tests are run with valgrind.

### DIFF
--- a/km/km_coredump.c
+++ b/km/km_coredump.c
@@ -1001,7 +1001,13 @@ int km_dump_core(char* core_path,
                         break;
                   }
                   strcat(buf, "\n");
-                  write(cfd, buf, strlen(buf));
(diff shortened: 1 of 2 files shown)


gnode1 (commented):
I thought we were doing away with the conf file, as it isn't needed with the new way of handling listening sockets.

paulpopelka:
> I thought we were doing away with the conf file, as it isn't needed with the new way of handling listening sockets.

We are.  I'm not sure how long that other review will take so we might want to put this in.
It is not a problem to discard this PR.

paulpopelka:
The change in PR #1719 removes the code where this fix is.
So, I am closing this PR.

PR #1721, opened 2023-01-19
Change the multilisten test to stop using expr

paulpopelka:
expr has a failure exit status if the expression it is processing evaluates to zero.
Which was causing bats to decide that the expr command failed.

### DIFF
--- a/tests/km_core_tests.bats
+++ b/tests/km_core_tests.bats
@@ -1713,14 +1713,14 @@ EOF
 
    # wait for the km mgmt pipe to exist
    tries=10
-   for ((i=0; i<10; i++))
+   while [ $tries -gt 0 ]
    do
       ls $MGTDIR | grep -q kmpipe
       if test $? -eq 0; then
          break;
       fi
       sleep 2
-      tries=`expr $tries - 1`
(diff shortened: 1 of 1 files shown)


gnode1 (approved):

tests/km_core_tests.bats
```diff
@@ -1713,14 +1713,14 @@ EOF
 
    # wait for the km mgmt pipe to exist
    tries=10
-   for ((i=0; i<10; i++))
+   while [ $tries -gt 0 ]
```
sv641, line 1716: **for i in {1..10}; do done** is bash way.

sv641 (approved):

tests/km_core_tests.bats
paulpopelka: I want to see how many tries are remaining when we exit the loop.
If $tries > 0 then the test succeed.  If $tries == 0 then the test failed.

I want the value of $tries to have a scope outside the loop.
I don't know what the scope of $i is in your example and I suspect others don't know either.
What I have is pretty clear as to the scope of $tries.

PR #1722, opened 2023-01-23
Don't run the bats multilisten-snapshot test with valgrind

paulpopelka:
valgrind needs the use of the km --putenv= parameter and snapshots do not allow adding variables to the payload's environment so the multilisten test was failing with this error:

09:15:39.636920 main                 636  memchec cannot set new environment when resuming a snapshot


gnode1:
Superseded by https://github.com/kontainapp/km/pull/1723

PR #1725, opened 2023-01-30
Create snapshots with holes where unbacked pages are in the dumped memory

paulpopelka:
To create a snapshot with unbacked memory pages put into the core file as holes, define SPARSE_COREFILE in the km environment when starting up km.  Currently we only produce sparse snapshot files.  It is easy to produce sparse corefiles if we think we need to.

This commit is mostly for people to look at the code.  There is no bats test yet.  There is some code added to the multilisten test to allocate 100MB with malloc() and 100MB with mmap().  malloc() creates backed pages, mmap() creates unbacked memory pages.

### DIFF
--- a/km/km_coredump.c
+++ b/km/km_coredump.c
@@ -43,6 +43,8 @@
 // TODO: Need to figure out where the corefile and snapshotdefault should go.
 static char* coredump_path = "./kmcore";
 
+int km_sparse_corefile = 0;   // if true, don't write unbacked data pages to the core file
+
 void km_set_coredump_path(char* path)
 {
    km_infox(KM_TRACE_COREDUMP, "Setting coredump path to %s", path);
@@ -586,6 +588,24 @@ static inline int km_core_write_notes(km_vcpu_t* vcpu,
    return 0;
 }
 
+static inline int km_corefile_align(int corefd, off_t* offp)
+{
+   off_t off = lseek(corefd, 0, SEEK_CUR);
+   if (off == (off_t)-1) {
+      km_warn("lseek( SEEK_CUR ) failed");
+      return errno;
+   }
+   if (off != roundup(off, KM_PAGE_SIZE)) {
+      off = roundup(off, KM_PAGE_SIZE);
+      if (lseek(corefd, off, SEEK_SET) == (off_t)-1) {
+         km_warn("lseek to %ld failed", off);
+         return errno;
+      }
+   }
+   *offp = off;
+   return 0;
+}
+
 /*
  * Write a contiguous area in guest memory. Since write can
  * be very large, break it up into reasonably sized pieces (1MB).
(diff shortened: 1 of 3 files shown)


km/km_coredump.c
```diff
@@ -626,6 +639,157 @@ static inline int km_guestmem_write(int fd, km_gva_t base, size_t length)
...
+   off_t pagemap_offset = ((uint64_t)kma_base / KM_PAGE_SIZE * PROC_PAGEMAP_ENTRYSIZE);
+   if (lseek(pagemapfd, pagemap_offset, SEEK_SET) < 0) {
+      km_warn("pagemap seek to 0x%lx failed, %s", pagemap_offset, strerror(errno));
+      close(pagemapfd);
+      free(pagemap);
+      return errno;
+   }
+   size_t bytestoread = numberofpages * PROC_PAGEMAP_ENTRYSIZE;
+   ssize_t bytesread = read(pagemapfd, pagemap, bytestoread);
+   if (bytesread < 0) {
```
sv641, line 714: I feel that there needs a check if required number of bytes are read.

km/km_coredump.c
```diff
@@ -626,6 +639,157 @@ static inline int km_guestmem_write(int fd, km_gva_t base, size_t length)
...
+   }
+
+   km_infox(KM_TRACE_COREDUMP, "base=0x%lx length=0x%lx file offset=0x%lx", base, length, base_offset);
+
+   // !!!! use km virtual address
+   km_kma_t kma_base = km_gva_to_kma(base);
+
+   // Get virtual memory page backing info for the passed memory range from /proc/self/pagemap
+   size_t bytestowrite = roundup(length, KM_PAGE_SIZE);
+   size_t numberofpages = bytestowrite / KM_PAGE_SIZE;
```
sv641, line 697: one more entry is needed. This is truncation.

km/km_coredump.c
```diff
@@ -586,6 +588,24 @@ static inline int km_core_write_notes(km_vcpu_t* vcpu,
    return 0;
 }
 
+static inline int km_corefile_align(int corefd, off_t* offp)
+{
+   off_t off = lseek(corefd, 0, SEEK_CUR);
```
sv641, line 593: This is a question not a comment. 
If the off is unaligned we are skipping the current page. Is this what we want?

km/km_coredump.c
```diff
@@ -626,6 +639,157 @@ static inline int km_guestmem_write(int fd, km_gva_t base, size_t length)
...
+
+   km_infox(KM_TRACE_COREDUMP, "base=0x%lx length=0x%lx file offset=0x%lx", base, length, base_offset);
+
+   // !!!! use km virtual address
+   km_kma_t kma_base = km_gva_to_kma(base);
+
+   // Get virtual memory page backing info for the passed memory range from /proc/self/pagemap
+   size_t bytestowrite = roundup(length, KM_PAGE_SIZE);
+   size_t numberofpages = bytestowrite / KM_PAGE_SIZE;
+   uint64_t* pagemap = malloc(numberofpages * PROC_PAGEMAP_ENTRYSIZE);
```
sv641, line 698: missing null pointer check

km/km_coredump.c
```diff
@@ -626,6 +639,157 @@ static inline int km_guestmem_write(int fd, km_gva_t base, size_t length)
...
+   int pagemapfd = open(PROC_PAGEMAP, O_RDONLY);
```
sv641, line 699: This function gets called several times. open/close can be moved out. But that is just an opinion.

sv641 (approved):
memory can change between calls to pagemap and write of memory.

paulpopelka:
> memory can change between calls to pagemap and write of memory.

All vcpu's are paused while a snapshot is taken.
So, unless km itself alters the payload, I don't expect the payloads memory to change.

sv641:
> 

When KM is proxy'ing systemcall and the snapshot happens we cannot predict what happens.
(i.e) we interrupt a systemcall and then pending IO's will still complete and page state will change.
This asynchronous behavior cannot be predicted.

paulpopelka:
> > 
> 
> When KM is proxy'ing systemcall and the snapshot happens we cannot predict what happens. (i.e) we interrupt a systemcall and then pending IO's will still complete and page state will change. This asynchronous behavior cannot be predicted.

The km will disallow snapshots while there are active asynch i/o contexts.

paulpopelka:
Closing this as it didn't provide much benefit.

PR #1726, opened 2023-02-03
setgroups, fix logic with size == 0

gnode1:
After this morning nginx conversation I tried it. The issue was with getgroups incorrectly checking for `list == NULL`. In fact it is allowed to be NULL if size is 0. 

>  If size is zero, list is not modified, but the total number of supplementary group IDs for the process is returned.  This allows the caller to  determine the size of a dynamically allocated list to be used in a further call to getgroups().

nginx runs OK with this fix


johnmuth81 (approved):
lgtm

gnode1:
My mistake on stating that nginx runs with this fix. Disregard that part.

PR #1727, opened 2023-02-08
Add support for timerfd syscalls, fix sendmsg() logging.

paulpopelka:
Add support for setuid()/setgid() syscalls.
Using serge's implementation from another in progress changeset of his.

I looked at a km trace that serge got from running ngnix under km. We saw several things and this changes addresses some of them. We saw trying to add an fd that was already in use.  This seems to come from sendmsg/recvmsg using SCM_RIGHTS to pass fd's between processes.
I also see that nginx tries to use the timerfd_create() family of syscalls, so added those.
nginx also tries to use io_setup() et al syscalls.  I've added the stubbed out functions to support this but implementation is needed.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -1800,25 +1800,37 @@ uint64_t km_fs_sendrecvmsg(km_vcpu_t* vcpu, int scall, int sockfd, struct msghdr
    }
    if (scall == SYS_sendmsg) {
       // translate sent file descriptors if any
-      for (struct cmsghdr* cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg)) {
-         if (cmsg->cmsg_type == SCM_RIGHTS) {
-            int guest_fd = *(int*)CMSG_DATA(cmsg);
-            int host_fd = km_fs_g2h_fd(guest_fd, NULL);
-            *(int*)CMSG_DATA(cmsg) = host_fd;
-            km_infox(KM_TRACE_FILESYS, "send guest fd %d as host %d", guest_fd, host_fd);
+      if (msg->msg_control != NULL) {
+         for (struct cmsghdr* cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg)) {
+            if (cmsg->cmsg_level == SOL_SOCKET && cmsg->cmsg_type == SCM_RIGHTS) {
+               int fdcount = (cmsg->cmsg_len - sizeof(struct cmsghdr)) / sizeof(int);
+	       km_infox(KM_TRACE_FILESYS, "sockfd %d: processing SCM_RIGHTS send chunk containing %d fd's", sockfd, fdcount);
+               int* guest_fd = (int*)CMSG_DATA(cmsg);
+	       for (int i = 0; i < fdcount; i++) {
+                  int host_fd = km_fs_g2h_fd(guest_fd[i], NULL);
+                  km_infox(KM_TRACE_FILESYS, "fd[%d] send guest fd %d as host fd %d", i, guest_fd[i], host_fd);
+                  guest_fd[i] = host_fd;
+               }
+            }
          }
       }
    }
    ret = __syscall_3(scall, host_sockfd, (uintptr_t)msg, flag);
-   if (scall == SYS_recvmsg) {
+   if (scall == SYS_recvmsg && ret >= 0) {
       // receive file descriptors if any
-      for (struct cmsghdr* cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg)) {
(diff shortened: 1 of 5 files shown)


km/km_filesys.c
```diff
@@ -1800,25 +1800,37 @@ uint64_t km_fs_sendrecvmsg(km_vcpu_t* vcpu, int scall, int sockfd, struct msghdr
...
+      if (msg->msg_control != NULL) {
+         for (struct cmsghdr* cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg)) {
+            if (cmsg->cmsg_level == SOL_SOCKET && cmsg->cmsg_type == SCM_RIGHTS) {
+               int fdcount = (cmsg->cmsg_len - sizeof(struct cmsghdr)) / sizeof(int);
+	       km_infox(KM_TRACE_FILESYS, "sockfd %d: processing SCM_RIGHTS send chunk containing %d fd's", sockfd, fdcount);
+               int* guest_fd = (int*)CMSG_DATA(cmsg);
+	       for (int i = 0; i < fdcount; i++) {
+                  int host_fd = km_fs_g2h_fd(guest_fd[i], NULL);
+                  km_infox(KM_TRACE_FILESYS, "fd[%d] send guest fd %d as host fd %d", i, guest_fd[i], host_fd);
+                  guest_fd[i] = host_fd;
```
paulpopelka, line 1812: This is altering the payloads memory but the fd isn't changing unless valgrind is being used.

We should use our own buffer for the altered control message.

km/km_filesys.c
```diff
@@ -1800,25 +1800,37 @@ uint64_t km_fs_sendrecvmsg(km_vcpu_t* vcpu, int scall, int sockfd, struct msghdr
...
+         for (struct cmsghdr* cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg)) {
+            if (cmsg->cmsg_level == SOL_SOCKET && cmsg->cmsg_type == SCM_RIGHTS) {
+               int fdcount = (cmsg->cmsg_len  - sizeof(struct cmsghdr)) / sizeof(int);
+	       km_infox(KM_TRACE_FILESYS, "sockfd %d: processing SCM_RIGHTS recv chunk containing %d fd's", sockfd, fdcount);
+               int* host_fd = (int*)CMSG_DATA(cmsg);
+	       for (int i = 0; i < fdcount; i++) {
+                  int guest_fd =
+                      km_add_guest_fd_internal(vcpu, host_fd[i], NULL, flag, KM_FILE_HOW_RECVMSG, NULL);
+                  km_infox(KM_TRACE_FILESYS, "fd[%d]: received host fd %d will be guest fd %d", i, host_fd[i], guest_fd);
+                  host_fd[i] = guest_fd;
```
paulpopelka, line 1831: This is also altering the payloads memory.
km should use its own buffer.

km/km_filesys.c
gnode1: I don't follow the valgrind reference here. Files are always mapped one to one. Memory is mapped differently with valgrind

paulpopelka: You are right, I'm not thinking clearly.

gnode1 (approved):
LGTM. Maybe remove the #if 0 ...

km/km_hcalls.c
```diff
@@ -1860,6 +1890,94 @@ static km_hc_ret_t getitimer_hcall(void* vcpu, int hc, km_hc_args_t* arg)
    return HC_CONTINUE;
 }
 
+/*
+ * int timerfd_create(int clockid, int flags);
+ */
+static km_hc_ret_t timerfd_create_hcall(void* vcpu, int hc, km_hc_args_t* arg)
+{
+   arg->hc_ret = km_fs_timerfd_create(vcpu, arg->arg1, arg->arg2);
```
sv641, line 1898: I think the pointer needs conversion

sv641, line 1898: my mistake

km/km_filesys.c
```diff
@@ -2284,6 +2306,15 @@ uint64_t km_fs_epoll_ctl(km_vcpu_t* vcpu, int epfd, int op, int fd, struct epoll
...
+{
+   int ret = __syscall_2(SYS_timerfd_create, clockid, flags);
+   if (ret >= 0) {
+      ret = km_add_guest_fd_internal(vcpu, ret, NULL, flags, KM_FILE_HOW_TIMERFD, NULL);
+   }
+   return ret;
+}
+
 static inline int
 km_fs_event_check_errors(km_vcpu_t* vcpu, km_file_t* file, struct epoll_event* events, int maxevents)
```
sv641, line 2319: I am not thinking straight where is this getting called from?

sv641 (approved):

km/km_filesys.c
paulpopelka: Doesn't look like it is called from anywhere.

Kind of looks like it was meant to be used as part of snapshot eventfd recovery.

PR #1730, opened 2023-02-10
Add the aio io_*() syscalls to km and a simple test program

paulpopelka:
The io_setup(), io_submit(), io_getevents(), io_cancel(), and io_destroy() hypercalls are added to km with this change. The bats aio_test program is added to exercise the hypercalls. I can't figure out how to make io_cancel() work.  It seems like the io request being canceled must be active.  If the io has completed, I assume it is nolonger cancelable and it returns an error code that doesn't tell me the io request has completed.

Putting this out for people to look at and comment on.

### DIFF
--- a/km/km_hcalls.c
+++ b/km/km_hcalls.c
@@ -2123,6 +2123,7 @@ static km_hc_ret_t io_setup_hcall(void* vcpu, int hc, km_hc_args_t* arg)
  */
 static km_hc_ret_t io_submit_hcall(void* vcpu, int hc, km_hc_args_t* arg)
 {
+   int needed_iovecs = 0;
    struct iocb km_iocb_copy[arg->arg2];
    struct iocb* km_iocb_list[arg->arg2];
    struct iocb** local_iocbpp = km_gva_to_kma(arg->arg3);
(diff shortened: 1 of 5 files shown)


tests/aio_test.c
```diff
@@ -0,0 +1,190 @@
...
+   fprintf(stdout, "Using %d files for this test run\n", nf);
+
+   // create data for test file contents and open test files
+   // and setup async io control blocks.
+   for (int i = 0; i < nf; i++) {
+      // fill buffer
+      memset(&buffer[i * chunksize], i, chunksize);
+
+      // open files, pid may not be unique in a container
+      snprintf(filename, sizeof(filename), "/tmp/aio_file_%d_%d", getpid(), i);
```
paulpopelka, line 61: I don't think getpid() is good enoughto generate a unique name in a container.
Wish there was a uuid generator function in the c library.

paulpopelka:
Need to add code to disallow a snapshot while an active aio context is present.

km/km_hcalls.c
```diff
@@ -2085,6 +2086,121 @@ static km_hc_ret_t mlock_hcall(void* vcpu, int hc, km_hc_args_t* arg)
...
+}
+
+/*
+ * int syscall(SYS_io_destroy, aio_context_t ctx_id);
+ */
+static km_hc_ret_t io_destroy_hcall(void* vcpu, int hc, km_hc_args_t* arg)
+{
+   arg->hc_ret = __syscall_1(SYS_io_destroy, arg->arg1);
+   if (arg->hc_ret == 0) {
+      km_assert(__atomic_load_n(&km_io_context_count, __ATOMIC_SEQ_CST) > 0);
```
sv641, line 2168: is there a race condition between assert and decrement?

paulpopelka: Yes.  Potentially one thread could be creating an i/o context while another is destroying one.
We can't trust the assertion here.

paulpopelka: I take my reply back.  We are only testing to ensure that the active context count doesn't go negative.
I think it should never go negative.

paulpopelka:
Snapshot active i/o contexts is not simple..  The kernel generates the id that represents a context.
So, the context id we capture in a snapshot can't be used again unless by chance the kernel gives it to the payload again when the snapshot is resumed and a replacement io context is created.
We can keep a table where we map pre snapshot io context id's to post snapshot context id's and map old to new id's on the io syscall.  Would probably need to worry about the kernel re-using a pre-snapshot context id post snapshot.
Another thing to consider is snapshots of snapshots and multi-level context id mapping.

PR #1731, opened 2023-02-14
cleanup error messages

gnode1:
km_err() uses the first arg as argument to exit(), and always prints errno. pthread_* functions don't set errno, return error instead. Calls to km_err() in that case were incorrect


km/km_snapshot.c
```diff
@@ -158,11 +158,7 @@ static inline void km_ss_recover_memory(int fd, km_gva_t tbrk_gva, km_payload_t*
                               fd,
                               phdr->p_offset);
                if (m == MAP_FAILED) {
-                  km_err(errno,
-                         "snapshot mmap[%d]: vaddr=0x%lx offset=0x%lx",
-                         i,
-                         phdr->p_vaddr,
-                         phdr->p_offset);
+                  km_err(2, "snapshot mmap[%d]: vaddr=0x%lx offset=0x%lx", i, phdr->p_vaddr, phdr->p_offset);
```
paulpopelka, line 161: Is there a reason you did not use km_err_e() here?
Same question for the change below.

paulpopelka: km_err() does print the errno.  Not thinking again.

PR #1735, opened 2023-02-16
WIP: snapshot create and restore hooks

johnmuth81:
This is snapshot create and restore hooks implementation. Right now it only works with the snapshot hypercall.


km/km_hcalls.c
```diff
@@ -1139,17 +1139,10 @@ static km_hc_ret_t rt_sigprocmask_hcall(void* vcpu, int hc, km_hc_args_t* arg)
...
+   km_sigaction_t* oldact = km_gva_to_kma(arg->arg3);
 
-   if (arg->arg2 != 0 && (act = km_gva_to_kma(arg->arg2)) == NULL) {
-      arg->hc_ret = -EFAULT;
-      return HC_CONTINUE;
-   }
-   if (arg->arg3 != 0 && (oldact = km_gva_to_kma(arg->arg3)) == NULL) {
-      arg->hc_ret = -EFAULT;
-      return HC_CONTINUE;
-   }
```
paulpopelka, line 1152: Please don't remove this.
With this change if arg2 or arg3 are an invalid payload address km will pass NULL to the kernel with this system call.
The kernel will see the NULL and think the caller didn't supply these arguments and perform the system call as though the arguments were not supplied when the call should have failed with error EFAULT.

km/km_signal.c
```diff
@@ -1067,6 +1089,9 @@ int km_sig_snapshot_recover(char* buf, size_t length)
       km_warnx("km_nt_sighand_t size mismatch - old snapshot?");
       return -1;
    }
+   if (nt_sa->signo == NSIG) {
```
paulpopelka, line 1092: I think you want to compare against KM_SIGSNAPCREATE here.

km/km_snapshot.c
```diff
@@ -529,10 +545,30 @@ int km_snapshot_block(km_vcpu_t* vcpu)
    in_snapshot = 1;
    pthread_mutex_unlock(&snap_mutex);
    km_vcpu_pause_all(vcpu, ALL);   // Wait for everyone to get to the pause point.
+
+   for (int i = 0; i < KVM_MAX_VCPUS; i++) {
+      if (machine.vm_vcpus[i] == NULL) {
+         continue;
```
paulpopelka, line 551: Could we put a break statement here?
There should be no more vcpu's after this.

km/km.h
```diff
@@ -183,6 +191,7 @@ typedef struct km_vcpu {
    uint64_t dr_regs[4];   // remember the addresses we are watching and have written into
                           // the processor's debugging facilities in DR0 - DR3.
    gdb_vcpu_state_t gdb_vcpu_state;   // gdb's per thread (vcpu) state.
+   km_snap_state_t snap_state;
```
paulpopelka, line 194: Could you add a statement in km_vcpu_init() to set snap_state to its initial value.  It currently defaults to 0 (SNAP_STATE_RUNNING).

km/km.h
```diff
@@ -140,6 +140,14 @@ typedef enum {
    PAUSED       // Paused in km_vcpu_handle_pause
 } __attribute__((__packed__)) km_vcpu_state_t;
 
+typedef enum {
+   SNAP_STATE_RUNNING,
+   SNAP_STATE_PAUSED,
+   SNAP_STATE_RUNHOOK_CREATE,
+   SNAP_STATE_RUNHOOK_RESTORE,
+   SNAP_STATE_RUNHOOK_RESTORE_LIVE,
+} km_snap_state_t;
```
paulpopelka, line 149: Can you explain why these states are not part of km_vcpu_state_t with a comment here?

km/km_signal.c
johnmuth81: yep

km/km_snapshot.c
johnmuth81: is that true?

km/km.h
johnmuth81: ok

km/km_snapshot.c
johnmuth81: ok

km/km.h
johnmuth81: ok

johnmuth81:
> Are there any changes in tests/km_core_bats.sh to try out these changes?

No. `tests/snapshot_test.c` was changed to test the hooks, but there are no changes to the `.bat` file.

PR #1741, opened 2023-03-01
Add snapshot support for contexts created by io_setup().

paulpopelka:
The io_setup() syscall creates an asynch i/o context and returns it to the caller. The payload has no way to control what value for the context is returned.  For snapshot of asynch io to work we must allow to the payload to continue using the context it received before a snapshot was taken.  To accomplish this km generates its own io contexts and maps them to the ones created by the kernel.  Only km generated io contexts are passed to the payload.  So when a payload uses an io context km translates that value to the kernel value returned with io_setup() before forwarding asynch io requests to the kernel.

These km generated io contexts are saved in a new elf note in snapshot files so that the context can be restored when the snapshot is resumed and then continue to be used by the payload.
Snapshots are only allowed if there are no active i/o requests associated with io contexts.  If there is active asynch io, snapshots requests are faileded.  Once all completed asynch i/o operations are harvested with io_getevents() snapshots are again allowed.

Also added bats tests to verify that snapshoting asynch io contexts works.


johnmuth81 (approved):
looks reasonable. well done.

PR #1742, opened 2023-03-01
Add CFI directives to guest code. Allows stack trace through SYSCALL.

johnmuth81:
To use in GDB: `add-symbol-file $WROKSPACE/km/build/km/km_guest_asmcode.o -o 0x8000008000`

### DIFF
--- a/km/km_guest_asmcode.s
+++ b/km/km_guest_asmcode.s
@@ -158,7 +158,7 @@ km_hcargs:
     .global __km_syscall_handler
 __km_syscall_handler:
     .cfi_startproc
-    .cfi_register 16, 2  # old %rip is in %rcx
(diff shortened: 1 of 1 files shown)


sv641 (approved):

gnode1 (approved):
not that I understand it fully, but looks OK

johnmuth81:
> not that I understand it fully, but looks OK

FYI. The `.cfi_startproc` and `.cfi_endproc` ensure there is a DWARF record for the function so GDB knows about it. This is just good hygyne. The important one is the `.cfi_register` which tells the DWARF system that `%rip` for the last 'caller' is in `%rcx` thanks to the `syscall` instruction.

PR #1743, opened 2023-03-01
Fix km to allow gdb to use the symbols in the vdso pages

paulpopelka:
Don't assign a filename to the vvar and vdso regions in the km memory busy map. km was assigning the non-filenames [vvar] and [vdso] to the vdso regions which apparently prevented gdb from checking to see if they were the vdso area.

A quick way to test is run gdb on a payload coredump and then run the "disassem gettimeofday" command.
You will see symbols like __vdso_gettimeofday in the output which tells you gdb was able to access the symbols stored in the vdso pages.

### DIFF
--- a/km/km_mem.c
+++ b/km/km_mem.c
@@ -348,15 +348,21 @@ static void km_add_vvar_vdso_to_guest_address_space(void)
       }
    }
 
+   /*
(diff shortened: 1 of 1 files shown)


sv641 (approved):

johnmuth81 (approved):
do less, get more. I like it.

PR #1748, opened 2023-04-06
WIP: sendmmsg support

johnmuth81:
Do not commit yet. A bit more work to do.


km/km_hcalls.c
```diff
@@ -322,6 +322,126 @@ static km_hc_ret_t sendrecvmsg_hcall(void* vcpu, int hc, km_hc_args_t* arg)
...
+   // *timeout);
+
+   if (arg->arg2 == 0) {
+      arg->hc_ret = -EFAULT;
+      return HC_CONTINUE;
+   }
+   struct mmsghdr* guest_mmsghdr = km_gva_to_kma(arg->arg2);
+
+   // Can't have more then UIO_MAXIOV mmsghdr's
+   if (arg->arg3 == 0 || arg->arg3 >= UIO_MAXIOV) {
```
paulpopelka, line 381: Is vlen limited to UIO_MAXIOV - 1?
The man page says:
"The value specified in vlen is capped to UIO_MAXIOV (1024)."
Kind of feels like UIO_MAXIOV is the largest allowed value.

From kernel source in net/socket.c:
In __sys_sendmmsg()
```
        if (vlen > UIO_MAXIOV)
                vlen = UIO_MAXIOV;
```
It's not even a failure, it just silently drops the excess.

km/km_hcalls.c
```diff
@@ -322,6 +322,126 @@ static km_hc_ret_t sendrecvmsg_hcall(void* vcpu, int hc, km_hc_args_t* arg)
...
+{
+   // int sendmmsg(int sockfd, struct mmsghdr *msgvec, unsigned int vlen, int flags);
+   // int recvmmsg(int sockfd, struct mmsghdr *msgvec, unsigned int vlen, int flags, struct timespec
+   // *timeout);
+
+   if (arg->arg2 == 0) {
+      arg->hc_ret = -EFAULT;
+      return HC_CONTINUE;
+   }
+   struct mmsghdr* guest_mmsghdr = km_gva_to_kma(arg->arg2);
```
paulpopelka, line 378: Check to see if the mmsghdr address was valid?

tests/getaddrinfo_test.c
```diff
@@ -0,0 +1,29 @@
...
+#include <stdio.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include "greatest/greatest.h"
+
+TEST test_huggingface()
+{
+   struct addrinfo* res;
+   int rc = getaddrinfo("huggingface.co", "443", NULL, &res);
```
paulpopelka, line 12: We are depending on getaddrinfo() to use sendmmsg() and recvmmsg() to test these interfaces?
So if they change getaddrinfo() at some point in the future these interfaces aren't being tested.

km/km_hcalls.c
johnmuth81: Yep. You're right. Changed to match kernel behavior.

km/km_hcalls.c
johnmuth81: Yep. That's a good idea. Added.

tests/getaddrinfo_test.c
johnmuth81: Yes I was, so I added a new test to exercise `sendmsg`, `recvmsg`, `sendmmsg`, and `recvmmsg` explicitly due to this well deserved chiding. Please see `tests/sendmsg_test.c`.

tests/sendmsg_test.c
```diff
@@ -0,0 +1,141 @@
...
+   char buffer2[1024];
+   memset(buffer2, 0, sizeof(buffer2));
+   struct iovec riov2 = {.iov_base = buffer2, .iov_len = sizeof(buffer2)};
+   struct mmsghdr rmmsghdr[2] = {
+                                {.msg_hdr = {.msg_iov = &riov, .msg_iovlen = 1}},
+                                {.msg_hdr = {.msg_iov = &riov2, .msg_iovlen = 1}}
+                                };
+
+   ASSERT(recvmmsg(recvfd, rmmsghdr, 2, 0, NULL) == 2);
+
```
paulpopelka, line 95: The sendmsg/recvmsg test validates the message contents but that validation is not done here.

PR #1754, opened 2023-05-12
Fix some uninitialized fields in the msghdr structure

paulpopelka:
valgrind found some problems in km sendmsg() code.

### DIFF
--- a/km/km_hcalls.c
+++ b/km/km_hcalls.c
@@ -305,6 +305,7 @@ static km_hc_ret_t sendrecvmsg_hcall(void* vcpu, int hc, km_hc_args_t* arg)
       iov[i].iov_len = iov_kma[i].iov_len;
    }
    msg.msg_iov = iov;
+   msg.msg_control = msg_kma->msg_control;
    // If msg_controlen is zero the msg_control pointer is not validated.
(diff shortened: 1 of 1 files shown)


km/km_hcalls.c
```diff
@@ -345,15 +346,16 @@ static int copyin_msghdr(struct msghdr* km_msg, struct msghdr* guest_msg)
    if (km_msg->msg_iovlen > 0) {
       for (int i = 0; i < km_msg->msg_iovlen; i++) {
          // Don't validate addresses if the element has a length of zero.
+         km_msg->msg_iov[i].iov_len = guest_msg->msg_iov[i].iov_len;
```
johnmuth81, line 349: `guest_msg->msg_iov` is a `km_gva_t`. Need to translate to kma before dereferencing. That's why the valgrind test is crashing since is doesn't use GVA == KMA.

johnmuth81 (changes requested):

paulpopelka:
Fix #1755 also fixes this and passes valgrind.
So abandoning this change.

johnmuth81:
These are actually better than #1755. Reopening.

johnmuth81 (approved):

PR #1755, opened 2023-05-12
fix valgrind uninitalized data error in sendmsg/recvmsg

johnmuth81:


km/km_hcalls.c
```diff
@@ -268,7 +268,7 @@ static km_hc_ret_t sendrecvmsg_hcall(void* vcpu, int hc, km_hc_args_t* arg)
    // ssize_t sendmsg(int sockfd, const struct msghdr *msg, int flags);
    // ssize_t recvmsg(int sockfd, struct msghdr *msg, int flags);
    struct msghdr* msg_kma = km_gva_to_kma(arg->arg2);
-   struct msghdr msg;
+   struct msghdr msg = {};
```
paulpopelka, line 271: I can't add a comment for copyin_msghdr so putting it here.
Could you add a note to the block comment about copyin_msghdr() that the caller must clear what km_msg points to before calling copyin_msghdr().

paulpopelka (approved):
I'll cancel my change.

johnmuth81:
Close in favor of #1774 (paul's)

PR #1756, opened 2023-05-14
check for zero size dup data, skip malloc

gnode1:
Noticed zero sized malloc on some snapshot restarts.


johnmuth81 (approved):
not much to it, eh?

gnode1:
No, but zero size malloc didn't feel right.

PR #1759, opened 2023-06-08
add glibc symbol support

sv641:
the code is very simple like paul said.

### DIFF
--- a/km/km_elf.h
+++ b/km/km_elf.h
@@ -26,8 +26,14 @@
 #include <stdint.h>
 #include <sys/mman.h>
 
-#define KM_DLOPEN_SYM_NAME "dlopen"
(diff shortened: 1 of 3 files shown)


paulpopelka (commented):
Could you try out the bats test: gdb_sharedlib for the glibc_dynamic test run?

Looks good.

gnode1:
>

gnode1:
> Could you try out the bats test: gdb_sharedlib for the glibc_dynamic test run?
> 
> Looks good.

I tried that. gdb_sharedlib test fails somewhere deep inside of our test infrastructure. We should fix it but right now it feels like a lot of work, for me at least.

I did running things manually, including c++ tests - all looks good.

gnode1 (approved):

PR #1766, opened 2023-07-05
Srini/gnu

sv641:
uniform usage of _GNU_SOURCE.

### DIFF
--- a/tests/sigaltstack_test.c
+++ b/tests/sigaltstack_test.c
@@ -178,7 +178,7 @@ GREATEST_MAIN_DEFS();
 
 int main(int argc, char** argv)
 {
-   printf("=== %d %d\n", SIGSTKSZ, MINSIGSTKSZ);
+   printf("=== %ld %ld\n", SIGSTKSZ, MINSIGSTKSZ);
 
    GREATEST_MAIN_BEGIN();   // init & parse command-line args
    greatest_set_verbosity(1);
--- a/tests/sigaltstack_test.c
+++ b/tests/sigaltstack_test.c
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <elf.h>
 #include <errno.h>
 #include <limits.h>
 #include <pthread.h>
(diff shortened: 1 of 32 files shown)


sv641:
with musl both SIGSTKSZ, MINSIGSTKSZ are -1 when compiled with _GNU_SOURCE
To get sigaltstack_test to pass added AT_MINSIGSTKSZ to auxv in km(which is missing from what kernel provides when starting new process)

paulpopelka:
I didn't understand what is going on here so I looked at the code for a while.

The real problem is the are 2 definitions of SIGSTKSZ and MINSIGSTKSZ.  Which definition you get depends on whether you define _GNU_SOURCE or not.
When we turned on _GNU_SOURCE you get the definition that calls sysconf (_SC_SIGSTKSZ) and the musl version of sysconf() does not support the conf type _SC_SIGSTKSZ so sysconf() fails and returns -1.
But no one apparent expects the SIGSTKSZ or MINSIGSTKSZ macro to fail.  :-)  Dontcha just love it.

So, to get around this problem we added the AT_MINSIGSTKSZ entry to the auxv that km builds for the payload.  Then we changed the sigaltstack test to call getauxval(AT_MINSIGSTKSZ) when the sysconf() call fails.
So that is how we now decide what size signal stack to use.

What a twisted web we weave.

tests/sigaltstack_test.c
```diff
@@ -178,7 +183,20 @@ GREATEST_MAIN_DEFS();
...
-   printf("=== %d %d\n", SIGSTKSZ, MINSIGSTKSZ);
+   /*
+    * musl does not support MINSIGSTKSZ and SIGSTKSZ
+    * when compiled with _GNU_SOURCE
+    */
+   min_signal_stack_size = MINSIGSTKSZ;
+   signal_stack_size = SIGSTKSZ;
+
+   if (min_signal_stack_size <= 0) {
+      min_signal_stack_size = getauxval(AT_MINSIGSTKSZ);
```
paulpopelka, line 194: You should fail the test if the call to getauxval() fails.

We got to this point because we weren't validating what SIGSTKSZ and MINSIGSTKSZ were returning.

sv641: added assertion

paulpopelka (approved):

PR #1768, opened 2023-07-18
add tag for vvar and vdso regions, this will stop MADV_DONTNEED

sv641:
vvar and vdso are always needed.


sv641:
> 

It is a longer fix. I will do that.

PR #1771, opened 2023-08-18
Fix faults caused by large reductions in the break size

paulpopelka:
We found problems when shrinking the break (brk() system call) from 1.5gig down to .5gig.
It looks like the page tables were not being updated properly when this happened.
This change hopefully fixes this problem and adds some tests to verify that memory is or is not accessible after the break is shrunk.


km/km_mem.c
```diff
@@ -262,6 +262,14 @@ static inline int PDE_SLOT(km_gva_t __addr)
 {
    return (__addr >> 21) & 0x1ff;
 }
+static inline int PDE_SLOT_ROUNDUP(km_gva_t __addr)
```
gnode1, line 265: We use roundup and rounddown functions elsewhere - is there a reason for these>

gnode1 (approved):

sv641 (approved):
This code looks good.

PR #1772, opened 2023-08-29
use the correct syscall return types

gnode1:
We got bitten by lseek to the end of 7GB file. I went through the file to check for other calls.

### DIFF
--- a/km/km_filesys.c
+++ b/km/km_filesys.c
@@ -907,7 +907,7 @@ uint64_t km_fs_prw(km_vcpu_t* vcpu, int scall, int fd, void* buf, size_t count, 
    if ((host_fd = km_fs_g2h_fd(fd, &ops)) < 0) {
       return -EBADF;
    }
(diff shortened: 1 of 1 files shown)


paulpopelka:
I assume this fixes the problem Sanjay was seeing in his testing?  The one he mentioned on the engineering slack channel.

gnode1:
> I assume this fixes the problem Sanjay was seeing in his testing? The one he mentioned on the engineering slack channel.

Yes. The issue he saw was with sleek. Others I just eyeballed.

gnode1:
> I assume this fixes the problem Sanjay was seeing in his testing? The one he mentioned on the engineering slack channel.

Many syscalls return int, I kept it that way. Uint64 is like a conduit, like void, to be cast accordingly on the other side.

paulpopelka (commented):
Seems like km_fs_getcwd() should get this treatment too.
I say this because the length argument to the syscall is a size_t.
I doubt this matters much since path length are never going to be 4gb or greater.

PR #1773, opened 2023-08-29
Change bats continue_after_snapshot test to discover if html has timed out

paulpopelka:
In the wonderful cloud environment our tests are treated to all manner of process scheduling preemptions.
In this case I thing the simple html server in the test was timed out by km_with_timeout before we finished the loop where we take snapshots and run curl against the html server.
We try to detect that km_cli or curl failed because the html server process timed out.  If that is what happened we end the test but call it successul.
If the html srver is still running then we assume there was some other failure and we do fail the test.

We will just need to wait for mr cloud to grace us with more kookie scheduling preemptions to see if this change helps.

### DIFF
--- a/tests/km_core_tests.bats
+++ b/tests/km_core_tests.bats
@@ -1655,8 +1655,17 @@ fi
    do
       #run ${KM_CLI_BIN} -r -d $SNAPDIR -s $MGTPIPE
       #assert_success
-      ${KM_CLI_BIN} -r -d $SNAPDIR -s $MGTPIPE
-      if test "$?" -ne 0; then
+      run ${KM_CLI_BIN} -r -d $SNAPDIR -s $MGTPIPE
+      if test "$status" -ne 0; then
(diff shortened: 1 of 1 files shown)


tests/km_core_tests.bats
```diff
@@ -1655,25 +1655,43 @@ fi
...
          sed -e "s/^/# /" $TRACEFILE >&3
          fail "km_cli failed, km trace is in the bats log"
       fi
       assert [ -f $SNAPDIR/kmsnap ]
 
       #run curl localhost:$socket_port
       #assert_success
-      curl localhost:$socket_port
-      if test "$?" -ne 0; then
+      run curl -4 localhost:$socket_port
```
sv641, line 1676: nitpicking - use --ipv4  easier to understand.

sv641 (approved):
Ship it.

gnode1 (approved):

PR #1779, opened 2023-09-01
Upgrade faktory net to 0.7.0. Force a dependabot change that aged out

johnmuth81:
This was a change that dependabot tried to apply and looks like it got lost somehow. Redo by hand to get rid of old alterts.


gnode1 (approved):

johnmuth81:
opps. Created from wrong base.

PR #1781, opened 2023-09-01
update dependabot config to weekly and target ecosystems

johnmuth81:


.github/dependabot.yml
```diff
@@ -5,8 +5,23 @@
 
 version: 2
 updates:
-  - package-ecosystem: "" # See documentation for possible values
+  - package-ecosystem: "gomod
```
gnode1, line 8: Is this right? Un-closed quotes?

johnmuth81: Doh! See update

gnode1 (approved):

PR #1786, opened 2023-09-04
Ez leka/action release

ez-leka:
- delendabot skip ci on pull-request
- release via github cli script


.github/dependabot.yml
```diff
@@ -9,4 +9,5 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
-
+    commit-message:
+      prefix: "[no ci] "
```
gnode1, line 13: Is that space at the end needed?

.github/workflows/release.yaml
```diff
@@ -83,7 +83,8 @@ jobs:
 
       - name: Build static krun
         run: |
-          make -C container-runtime static
+          # make -C container-runtime static
+          touch container-runtime/crun/krun.static
```
gnode1, line 87: ???

PR #1792, opened 2023-09-06
adding schedule run to release build

ez-leka:


.github/workflows/release.yaml
```diff
@@ -21,6 +21,16 @@ on:
...
+  schedule:
+    # This will run every Friday at 4 am and create draft v0.1-test
+    # Posix cron format:
+    # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07
+    # Minute Hour DayOfMonth MonthOfYear DayOfWeek
+    - cron: "0 4 * * 5" # Every friday at 4 am (not to overlap with CI nightly)
+    # Github doc says:
+    #    Scheduled workflows run on the latest commit on the default or base branch.
+    #    The shortest interval you can run scheduled workflows is once every 5 minutes.
+
```
gnode1, line 33: This will compile but AFAIK it won't build and publish artifacts. I think we need to somehow test that too

PR #1794, opened 2023-09-10
handle ioctl errno

gnode1:
Turns out modern KVM `ioctl (KVM_RUN)` _sometimes_ returns -errno and keeps proper errno untouched. According to `man ioctl()` it should return -1 and set `errno` in case of an error. However I observe that ioctl returning -14 (EFAULT) and errno is left untouched.

### DIFF
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -517,7 +517,8 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
               vcpu->regs.rsp,
               vcpu->sregs.cr2);
       vcpu->cpu_run->exit_reason = KVM_EXIT_INTR;
-      switch (ioctl_errno) {
+      errno = ioctl_errno;
+      switch (errno) {
          case EINTR:
             break;
 
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -491,12 +491,13 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
    int rc;
 
    vcpu->state = IN_GUEST;
-   errno = 0;
    vcpu->cpu_run->exit_reason = KVM_EXIT_UNKNOWN;   // Clear exit_reason from the preceding ioctl
    vcpu->regs_valid = 0;                            // Invalidate cached registers
    vcpu->sregs_valid = 0;
    km_infox(KM_TRACE_VCPU, "about to ioctl( KVM_RUN )");
+   errno = 0;
    rc = ioctl(vcpu->kvm_vcpu_fd, KVM_RUN, NULL);
+   int ioctl_errno = errno;
    vcpu->state = HYPERCALL;
    km_infox(KM_TRACE_VCPU,
             "ioctl( KVM_RUN ) returned %d KVM_RUN exit %d (%s)",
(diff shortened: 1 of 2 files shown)


johnmuth81 (approved):
KM is public, so I guess I still see it.

LGTM ;)

But I don't have write access, so my approval means nothing :(

km/km_vcpu_run.c
```diff
@@ -509,6 +510,9 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
    }
 
    if (rc != 0) {
+      if (errno == 0) {
```
paulpopelka, line 513: This does need a comment to the effect that we are compensating for some kind of bug in ioctl()'s ability to set errno properly.

sv641: errno most probably has changed between line 499 - 513(depending on what options we run with). we need to save it between lines 499 and 500.

paulpopelka (approved):
Just one minor comment.

And, github says this is being committed with my github id.  Is there a workspace on my machine that contains this?

km/km_vcpu_run.c
paulpopelka: Then this needs to be fixed differently.
Save errno after ioctl() returns and use that value to restore errno.

Plus ioctl() should not return -errno on failure but that is apparently what we see.

sv641: `int
__ioctl (int fd, unsigned long int request, ...)
{
  va_list args;
  va_start (args, request);
  void *arg = va_arg (args, void *);
  va_end (args);

  int r;
  if (!__ioctl_arch (&r, fd, request, arg))
    {
      r = INTERNAL_SYSCALL_CALL (ioctl, fd, request, arg);
      if (__glibc_unlikely (INTERNAL_SYSCALL_ERROR_P (r)))
	{
	  __set_errno (-r);
	  return -1;
	}
    }
  return r;
}`

paulpopelka, line 513: I think the proper fix is to save errno in a local variable  immediately after ioctl( KVM_RUN ) returns.
And then when we discover ioctl() has failed, restore errno so that km_trace() will display it properly and the switch statement below this will be operating on a valid value.

paulpopelka (commented):
I think you brought this fix over from km-gpu because it was thought to be some sort of problem with how ioctl() worked.
We have since found that the ioctl() hook we are using in km-gpu was causing the problem we are trying to fix here.
Since we don't hook ioctl() in this repository this fix does not apply here.

But, we do need to save errno since there are several lines of code between the call to ioctl( KVM_RUN ) and the point where we want to look at errno.  And some of those intervening lines can modify errno.

km/km_vcpu_run.c
gnode1: you right, will do

tests/km_core_tests.bats
```diff
@@ -1675,11 +1675,11 @@ fi
       #assert_success
       run curl -4 localhost:$socket_port
       if test "$status" -ne 0; then
-         run /bin/kill -0 $pid
+         run /bin/kill -0 $pid >& /dev/null
          if test $status -ne 0; then
             # html server is not running, if km_with_timeout timed it out, we give up on this loop and dont fail the test
-            wait -p kmstatus $pid
-            if test $kmstatus -eq 124; then
+            wait kmstatus $pid
```
paulpopelka, line 1681: Shouldn't this be " wait $pid" ?

km/km_vcpu_run.c
```diff
@@ -515,7 +517,7 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
               vcpu->regs.rsp,
               vcpu->sregs.cr2);
       vcpu->cpu_run->exit_reason = KVM_EXIT_INTR;
-      switch (errno) {
+      switch (ioctl_errno) {
```
paulpopelka, line 520: I think you should restore errno.
The invocation of km_err() in the default cause wants to use errno.

km/km_vcpu_run.c
```diff
@@ -495,7 +495,9 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
    vcpu->regs_valid = 0;                            // Invalidate cached registers
    vcpu->sregs_valid = 0;
    km_infox(KM_TRACE_VCPU, "about to ioctl( KVM_RUN )");
+   errno = 0;
```
paulpopelka, line 498: Is this really needed?

gnode1: ioctl doesn't set it if there is no error. So the code below would act on old value

PR #1798, opened 2023-09-25
Try to fix the ci failure in the bats gdb_nextstep test

paulpopelka:
We've been seeing this failure in the gdb_netstep test lately.
# 09:07:59.624655 km_gdb_set_thread_vc 1648 memchec Assertion `vcpu->state != IN_GUEST' failed: No such file or directory

The only recent change "seemingly" related to this was a change where we save the value of errno for use later in km_vcpu_one_kvm_run(). That change shifted the setting of state to IN_GUEST one statement further away from the ioctl( KVM_RUN ) call.

This change is more an experiment to see if it "helps" fix the assertion failure.  If it does "fix" the problem, we will know there is a race here that we've only made less likely but probably not fixed.

### DIFF
--- a/km/km_vcpu_run.c
+++ b/km/km_vcpu_run.c
@@ -490,15 +490,15 @@ static void km_vcpu_one_kvm_run(km_vcpu_t* vcpu)
 {
    int rc;
 
-   vcpu->state = IN_GUEST;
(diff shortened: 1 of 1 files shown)


paulpopelka:
Dam git, removed part of my checkin comment.

sv641 (approved):

paulpopelka:
This won't fix the problem.
Serge has identified a more likely cause.
Issue #1799 has been filed for the problem.

PR #1805, opened 2023-11-06
Change sendmsg_test to use assigned port numbers

paulpopelka:
It is currently using 2 fixed port numbers which can cause EADDRINUSE errors sometimes.
Chnaged sendmsg_test to use port numbers assigned it by the bats test script.


sv641 (approved):
missing freeaddrinfo

paulpopelka:
> missing freeaddrinfo

I don't see that.  Where are you seeing this?

PR #1808, opened 2023-11-21
Change aio_test to use bats primatives properly.

paulpopelka:


tests/km_core_tests.bats
```diff
@@ -1819,28 +1820,35 @@ EOF
       tries=$(($tries - 1))
    done
    assert [ $tries -ne 0 ]
-   rm -f $WORKDIR/waiting
+   run rm -f $WORKDIR/waiting
+   assert_success
```
gnode1, line 1824: Isn't `rm -f` always successful? Unless `waiting` might somehow become a directory.

I might misunderstand something here, but it seems to me there is no need to check this, nor to do `run rm`

tests/km_core_tests.bats
```diff
@@ -1819,28 +1820,35 @@ EOF
...
+   wait_and_check $aio_test_pid 0
+
+   # Get the snapshot file name
    rm -f $WORKDIR/kmsnap.*.conf
    snapfile=`ls $WORKDIR/kmsnap.aio_test$ext.[0-9]*`
-   assert_success
+   assert [ -n "$snapfile" ]
 
    # start the snapshot and tell it to get going
    run km_with_timeout $snapfile &
```
gnode1, line 1839: `run` shouldn't be here

tests/km_core_tests.bats
```diff
@@ -1860,8 +1868,8 @@ EOF
    run touch $WORKDIR/continue
```
gnode1, line 1868: Do we need `run` here? Feels unnecessary

tests/km_core_tests.bats
```diff
@@ -1860,8 +1868,8 @@ EOF
...
    assert_success
```
gnode1, line 1869: Not sure what we are checking here - can the test affect this touch somehow?

gnode1 (changes requested):

tests/km_core_tests.bats
paulpopelka: If touch fails the test will hang.
I'd like to fail the test rather that leave it hanging for 250 seconds and then have it timeout.

tests/km_core_tests.bats
paulpopelka: If touch fails, the test will hang.

tests/km_core_tests.bats
paulpopelka: The run is not needed, nor the assert_success

tests/km_core_tests.bats
paulpopelka: This run is also not needed.

tests/km_core_tests.bats
gnode1: why would it fail?

paulpopelka: I don't know.  It shouldn't fail but we are testing assumptions.
Maybe we could remove all the checks since the code shouldn't fail.

tests/km_core_tests.bats
```diff
@@ -1799,7 +1799,7 @@ EOF
 }
 
 @test "aio($test_type): exercise io_*() syscalls (aio_test$ext)" {
-   WORKDIR=aio_workdir.$$
+   WORKDIR=`pwd`/aio_workdir.$$
    mkdir -p $WORKDIR
    assert_success
```
gnode1, line 1804: pointless without run

tests/km_core_tests.bats
```diff
@@ -1822,25 +1823,31 @@ EOF
...
 
-   # wait for the snapshot to finish
+   # wait for aio_test snapshot to terminate
    wait_and_check $pid 0
 
    # Now try to take a snapshot when async i/o is active.
    # This snapshot should fail.
    rm -f $WORKDIR/*
    KM_MGTDIR=$WORKDIR  AIO_TEST_WORKDIR=$WORKDIR  run km_with_timeout aio_test$ext -f 13 -c 1 -sf &
```
gnode1, line 1849: run in backgroud

gnode1 (changes requested):

gnode1 (approved):

PR #1826, opened 2024-11-20
fedora 41 works now

sv641:


km/km_filesys.c
```diff
@@ -2273,7 +2273,7 @@ uint64_t km_fs_select(km_vcpu_t* vcpu,
 uint64_t km_fs_poll(km_vcpu_t* vcpu, struct pollfd* fds, nfds_t nfds, int timeout)
 {
    for (int i = 0; i < nfds; i++) {
-      if (km_fs_g2h_fd(fds[i].fd, NULL) < 0) {
+      if ((fds[i].fd != 0xffffffff) && (km_fs_g2h_fd(fds[i].fd, NULL) < 0)) {
```
paulpopelka, line 2276: The poll man page says a negative fd (not just -1) means that entry in the pollfd array should be ignored and that revent should be set to zero.

tests/signal_test.c
```diff
@@ -45,8 +45,7 @@ void* stack_addr = NULL;
 
 void signal_handler(int signal)
 {
-   char buf[128];
-   stack_addr = buf;
+   asm volatile("movq %%rsp, %0" : "=r"(stack_addr));
```
paulpopelka, line 48: Does this fix something or is it just a better way of doing things?
Just wondering.

km/km_filesys_private.h
```diff
@@ -60,7 +60,9 @@ typedef enum km_file_how {
    KM_FILE_HOW_SOCKETPAIR1 = 7,
    KM_FILE_HOW_RECVMSG = 8,
    KM_FILE_HOW_EVENTFD = 9, /* eventfd() */
-   KM_FILE_HOW_TIMERFD = 10
+   KM_FILE_HOW_TIMERFD = 10,
+   KM_FILE_HOW_WATCH = 11,
+   KM_FILE_HOW_WATCH1 = 12
```
paulpopelka, line 65: Would suggest you add code to km_fs_core_notes_write() to disallow snapshots of payloads that use inotify until we can support snapshoting inotify fd's.

km/km_hcalls.c
```diff
@@ -2209,17 +2209,30 @@ static km_hc_ret_t sched_yield_hcall(void* vcpu, int hc, km_hc_args_t* arg)
 
 static km_hc_ret_t inotify_init_hcall(void* vcpu, int hc, km_hc_args_t* arg)
 {
-   // int inotify_init(void);
-   km_warnx("Unsupported call inotify_init");
-   arg->hc_ret = -ENOTSUP;
+   arg->hc_ret = km_fs_inotify_init(vcpu);
```
paulpopelka, line 2212: Just curious why you didn't turn this into a call to km_fs_inotify_init1() and set the flags argument to zero.
What you have is fine though.

